PR #135, opened 2017-03-01
VTCCA-393 Return 400 and 404 for postcode lookups

MarkPaxton:
which are badly formatted or result in no data


app/controllers/AddressLookup.scala
```diff
@@ -25,7 +26,17 @@ trait AddressLookup extends PropertyLinkingController {
   val addresses = Wiring().addresses
 
   def findByPostcode(postcode: String) = Action.async { implicit request =>
-    addresses.findByPostcode(postcode) map { res => Ok(Json.toJson(res)) }
+    if (postcode.trim.contains(" ")) {
```
richard-dennehy: the postcode lookup should accept partial postcodes e.g. EX1 - and these currently work on CI3

test/controllers/AsyncControllerSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package controllers
+
+import org.scalatest.{AppendedClues, AsyncFlatSpec, BeforeAndAfterEach, MustMatchers}
+import play.api.test.DefaultAwaitTimeout
+import utils._
+
+trait AsyncControllerSpec extends AsyncFlatSpec with MustMatchers with DefaultAwaitTimeout with BeforeAndAfterEach with AppendedClues {
```
richard-dennehy: maybe I'm blind, but is this being used by anything?

app/controllers/AddressLookup.scala
MarkPaxton: Would search by post code prefix be so general as to be not useful? It would usually result in 100s of properties and that wouldn't be easy to choose from.

PR #186, opened 2017-04-28
remove old feature flags

richard-dennehy:


Nugs:
Can remove the whitelist block in application.conf as well.

PR #187, opened 2017-04-28
add controls to non-js pagination

richard-dennehy:


Nugs:
Can you update https://jira.tools.tax.service.gov.uk/browse/VTCCA-622 as done then as well please?

PR #198, opened 2017-05-12
No more keystore

chdupre:


app/controllers/agentAppointment/AppointAgentController.scala
```diff
@@ -102,7 +106,9 @@ class AppointAgentController extends PropertyLinkingController {
     val hasCheckAgent = pLink.agents.map(_.checkPermission).contains(StartAndContinue)
     val hasChallengeAgent = pLink.agents.map(_.challengePermission).contains(StartAndContinue)
     if (hasCheckAgent && agent.canCheck == StartAndContinue || hasChallengeAgent && agent.canChallenge == StartAndContinue) {
-      sessionRepository.start(agent, agentOrgId, pLink).map(_ => {
+      //  def start(agent: AppointAgent, agentOrgId: Long, propertyLink: PropertyLink)(implicit hc: HeaderCarrier): Future[Unit] = {
```
richard-dennehy: is this meant to still be here?

app/repositories/SessionRepository.scala
```diff
@@ -63,14 +70,23 @@ class SessionRepository @Inject()(db: DB)
       sessionId <- getSessionId
       maybeOption <- findById(sessionId)
     } yield {
-      maybeOption.map(_.data.as[A])
+      val aaa = maybeOption.map(_.data \ formId)
```
richard-dennehy: could we have a slightly more descriptive name than "aaa"?

PR #243, opened 2017-09-12
VTCCA-831 search and sort properties styling fixed and updates

peter-hazell:


darrell-plant:
search and sort properties styling fixed and updates

PR #298, opened 2017-10-24
Daniel/javascript pagination

DanielConnelly:
CSS still needs looked at to get the code looking like prototype.


DanielConnelly:
The css is not too far off prototype, it is just one line underneath where it should be struggled for near a full day getting it there.

darrell-plant:
To maintain consistency across search&sort and message tabs we will use the pagination implementation used on the messages tab instead of the solution on this branch.

PR #469, opened 2018-05-01
Vtcca 1721 delete draft confirm page

derekspalmer:


app/views/dashboard/confirmDeleteDraftCase.scala.html
```diff
@@ -0,0 +1,69 @@
...
+
+
+  <h2 class="heading-medium">@Messages("draftCases.confirm.delete.message.1")</h2>
+
+   @form(routes.ManageDrafts.confirmDelete(model.id)) {
+       <table id="draftsTable" class="margin-top-20 margin-bottom-20">
+        <caption class="visuallyhidden">@Messages("draftCases.confirm.table.caption")</caption>
+        <colgroup>
+         <col width="60%">
+         <col width="20%">
```
DanielConnelly: Is there a reason for the table stuff here? is it not just CSS we need to get right?

DanielConnelly: resolved

PR #514, opened 2018-06-07
VTCCA-1819 minor content fix

Fimartin1:


app/views/createAccount/groupAlreadyExists.scala.html
```diff
@@ -29,7 +29,7 @@ <h1 class="transaction-banner__heading">@Messages("groupAlreadyExists.title", bu
                   <p class="heading-medium no-margin">@Messages("groupAlreadyExists.addedToGroup")</p>
                 </div>
 
-                <p>@Messages("account.confirmation.disclaimer")</p>
+                <p>@Messages("account.confirmation.disclaimer.oldJourney")</p>
```
peter-hazell: I think this needs to be added to a couple of other files as well.. will show you in the morning @Fimartin1

PR #600, opened 2018-11-30
VTCCA-2332 DVR screens

DanielConnelly:


app/models/Accounts.scala
```diff
@@ -16,10 +16,11 @@
 
 package models
 
-import play.api.libs.json.Json
+import play.api.libs.json.{Json, Reads}
 
 case class Accounts(organisation: GroupAccount, person: DetailedIndividualAccount)
 
 object Accounts {
+  Reads
```
habeeb01: is the Reads needed?

conf/application.conf
```diff
@@ -46,6 +46,7 @@ featureFlags {
   fileUploadEnabled = false
   newDashboardRedirectsEnabled = true
   checkCasesEnabled = true
+  dvrEnabled = true
```
habeeb01: can we switch off the flag for now

PR #623, opened 2019-01-30
Vtcca attachment

DanielConnelly:


app/controllers/DvrController.scala
```diff
@@ -121,7 +121,7 @@ class DvrController @Inject()(
           dvrCaseManagement
             .getDvrDocument(link.uarn, valuationId, link.submissionId, fileRef)
             .map { document =>
-              Result(ResponseHeader(200, document.headers.updated(CONTENT_DISPOSITION, s"""attachment;filename="Untitled Document"""")),
+              Result(ResponseHeader(200, document.headers.updated(CONTENT_DISPOSITION, s"""attachment;filename="Untitled Document.pdf"""")),
```
PeterPerhac: is there really no better, meaningful file name? perhaps use `${link.submissionId}.pdf`

PeterPerhac:
as a user, i get really annoyed when downloading files and they all come down as "untitled.pdf" I think coming up with a meaningful name would make user experience **so much better**, also, using the submission id in the file name will avoid you ending up with downloads folder full of "foo.pdf, foo (1).pdf, foo (2).pdf"

PR #687, opened 2019-08-05
VTCCA-2759 15-minute timeout

reddy-yattapu:


app/controllers/KeepAliveController.scala
```diff
@@ -0,0 +1,32 @@
...
+import javax.inject.Inject
+
+import auth.VoaAction
+import play.api.i18n.{I18nSupport, MessagesApi}
+import uk.gov.hmrc.play.frontend.controller.FrontendController
+
+import scala.concurrent.Future
+
+class KeepAliveController @Inject()(val messagesApi: MessagesApi,
+                                    ggAction: VoaAction) extends FrontendController with I18nSupport {
```
DanielConnelly: we do not need I18nSupport

PR #706, opened 2019-09-03
VTCCA-2886 Bad request when attempting to remove an uploaded document…

reddy-yattapu:
… during Property Linking journey


jon-pike:
👍

habeeb01:
👍

PR #711, opened 2019-09-10
VTCCA-2292 Property Linking Pages showing duplicate h1 tags

reddy-yattapu:


app/views/main_template.scala.html
```diff
@@ -71,21 +71,13 @@
 }
 }
 
-@contentHeaderTextAsHtml = @{
```
DanielConnelly: what is this used for reddy before you go just deleting it?

habeeb01:
LGTM 👍

app/views/main_template.scala.html
reddy-yattapu: This code is outdated and commit as part of https://jira.tools.tax.service.gov.uk/browse/VTCCA-1232

DanielConnelly:
lgtm 👍

PR #722, opened 2019-10-08
VTCCA-2977: don't use authorised-frontend, use auth-client

PeterPerhac:


app/actions/GgAuthenticatedAction.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  val logger = Logger(this.getClass.getName)
+
+  override def invokeBlock[A](request: Request[A], block: RequestWithUserDetails[A] => Future[Result]): Future[Result] = {
+    implicit val hc: HeaderCarrier = fromHeadersAndSession(request.headers, Some(request.session))
+    logger.debug("the request called invoke block")
+    success(block)(request, hc)
+  }
+
+  def success[A](body: RequestWithUserDetails[A] => Future[Result])
```
DanielConnelly: as we are only using this in invoke block, is it not worth just putting it in invoke block?

app/actions/GgAuthenticatedAction.scala
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    val retrieval = name and email and postCode and groupIdentifier and externalId and affinityGroup and credentialRole
+    authorised(AuthProviders(GovernmentGateway)).retrieve(retrieval) {
+      case name ~ optEmail ~ optPostCode ~ Some(groupIdentifier) ~ Some(externalId) ~ Some(affinityGroup) ~ Some(role) =>
+        body(new RequestWithUserDetails(UserDetails.fromRetrieval(name, optEmail, optPostCode, groupIdentifier, externalId, affinityGroup, role), request))
+    }.recoverWith(handleError)
+  }
+
+  def noVoaRecord: Future[Result] =
```
DanielConnelly: is this used?

app/models/Nino.scala
```diff
@@ -0,0 +1,53 @@
+/*
```
DanielConnelly: can we put this in a domain package?

DanielConnelly: or even a registration, as we only use the nino in that case, I dont want the Nino getting mixed in with VOA models.

test/controllers/RegistrationControllerSpec.scala
```diff
@@ -182,16 +154,11 @@ class RegistrationControllerSpec extends VoaPropertyLinkingSpec with MockitoSuga
 
   "Going to the create account page when logged in as a new admin user registering with an existing group account" should
     "display the complete your contact details form for an admin" in {
+    val u: UserDetails = userDetails(affinityGroup = Organisation, credentialRole = Admin)
```
DanielConnelly: can we call this something better than u, like userdetails?

app/actions/GgAuthenticatedAction.scala
DanielConnelly: there are no tests for this.

DanielConnelly:
few comments, missing some tests for newly created stuff but over all very good clean up. (I believe we can go further but lets see it all working)

PR #723, opened 2019-10-08
VTCCA-2976 file upload issue rework with extra validation

DanielConnelly:


app/actions/AuthenticatedAction.scala
```diff
@@ -71,31 +68,36 @@ class AuthenticatedAction @Inject()(override val messagesApi: MessagesApi,
                   body: BasicAuthenticatedRequest[A] => Future[Result])
                 (implicit request: Request[A], hc: HeaderCarrier): Future[Result] = {
     def handleError: PartialFunction[Throwable, Future[Result]] = {
-      case _: NoActiveSession =>
+      case _: InsufficientEnrolments    =>
```
darrell-plant: This has the issue we saw the other day where the body will be executed twice.

PeterPerhac: no , i saw that when i was working on auth library stuff, and i believe I did fix the double-running of the body, it's somewhere lower down in this file

PeterPerhac:
that looks like the service is getting neater and slimmer by the minute! 👍 nice

darrell-plant:
👍  after code walk-through

PR #725, opened 2019-10-10
VTCCA credential role invalid exception not being caught, handling er…

DanielConnelly:
…ror and adding Admin credential role to potentially fill gap.


habeeb01:
👍 LGTM

PR #727, opened 2019-10-10
Vtcca credential role bug

DanielConnelly:


PeterPerhac:
👍 yeah, that sucks we have to do that, but order matters!

PR #734, opened 2019-10-30
VTCCA-3040: adding IDs as requested to aid testing

PeterPerhac:
20 new IDs added to a variety of elements on the property linking pages


darrell-plant:
LGTDz 👍

PR #736, opened 2019-11-01
VTCCA-3043: make sure agentAppointed field is send to the backend

habeeb01:


app/controllers/agentAppointment/AppointAgentController.scala
```diff
@@ -100,13 +100,17 @@ class AppointAgentController @Inject()(
...
-        request.organisationAccount.id, agentOrganisation.fold(throw new IllegalArgumentException("agent organisation required."))(_.id))
+      response <- agentRelationshipService.getMyOrganisationPropertyLinksWithAgentFiltering(
+        params = params,
+        pagination = AgentPropertiesParameters(
+          agentCode = agentCode,
+          checkPermission = AgentPermission.fromName(checkPermission).getOrElse(StartAndContinue),
+          challengePermission = AgentPermission.fromName(challengePermission).getOrElse(StartAndContinue),
+          pageNumber = pagination.page,
+          pageSize = pagination.pageSize,
+          agentAppointed = agentAppointed.getOrElse("BOTH")
```
darrell-plant: what is "BOTH"?

habeeb01: we default to BOTH because that will bring back all of the user's properties

habeeb01: BOTH is the default on the API as well

darrell-plant: then `AgentPropertiesFilter.Both.name` ?

habeeb01: updated

darrell-plant:
LGTM 👍

jon-pike:
👍

PR #738, opened 2019-11-06
VTCCA-3023: make sure the persisted evidenceType is reloaded in the form

habeeb01:


jon-pike:
👍

app/controllers/propertyLinking/UploadController.scala
```diff
@@ -43,7 +43,7 @@ class UploadController @Inject()(
                                   val errorHandler: CustomErrorHandler,
                                   authenticatedAction: AuthenticatedAction,
                                   withLinkingSession: WithLinkingSession,
-                                  businessRatesAttachmentsServices: BusinessRatesAttachmentService
+                                  businessRatesAttachmentsService: BusinessRatesAttachmentService
```
PeterPerhac: i think now the name of the parameter is good `businessRatesAttachmentsService` but the name of the service should be plural "Attachments" not "Attachment" (BusinessRatesAttachment**s**Service)

habeeb01: cool, will do that. Will do the same for the connector as well

reddy-yattapu:
:+1:

PR #741, opened 2019-11-21
VTCCA-2980 adding feature flag for new external-valuation-management-…

DanielConnelly:
…api call

### DIFF
--- a/app/config/GuiceModule.scala
+++ b/app/config/GuiceModule.scala
@@ -20,11 +20,15 @@ import java.time.Clock
 
 import com.google.inject.AbstractModule
 import com.google.inject.name.Names
+import com.google.inject.name.Names.named
+import com.typesafe.config.ConfigException
 import play.api._
 import repositories._
 import services._
 import services.iv.{IdentityVerificationService, IvService}
 import uk.gov.hmrc.play.bootstrap.config.{RunMode, ServicesConfig}
+
+import scala.util.Try
 
 class GuiceModule(
                    environment: Environment,
@@ -32,6 +36,8 @@ class GuiceModule(
                  ) extends AbstractModule {
 
   def configure() = {
+
+    bindBoolean("feature.externalValuation", "detailed-valuation.external")
 
     bind(classOf[ServicesConfig]).toInstance(new ServicesConfig(configuration, new RunMode(configuration, environment.mode)))
     bind(classOf[SessionRepo]).annotatedWith(Names.named("propertyLinkingSession")).to(classOf[PropertyLinkingSessionRepository])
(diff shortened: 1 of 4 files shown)


app/config/GuiceModule.scala
```diff
@@ -42,4 +48,15 @@ class GuiceModule(
     bind(classOf[Clock]).toInstance(Clock.systemUTC())
   }
 
+  protected def bindBoolean(path: String, name: String = ""): Unit =
+    bindConstant()
+      .annotatedWith(named(resolveAnnotationName(path, name)))
+      .to(Try(configuration.get[String](path).toBoolean).toOption.getOrElse(configException(path))) //We need to parse as string, due to the process of adding in from app-config-<env> it is seen as a string
+
+  private def resolveAnnotationName(path: String, name: String): String = name match {
```
darrell-plant: I don't know why we bother with `resolveAnnotationName`. Isn't it better to always provide a name?

darrell-plant:
LGTM 👍

app/config/GuiceModule.scala
```diff
@@ -42,4 +48,15 @@ class GuiceModule(
     bind(classOf[Clock]).toInstance(Clock.systemUTC())
   }
 
+  protected def bindBoolean(path: String, name: String = ""): Unit =
+    bindConstant()
+      .annotatedWith(named(resolveAnnotationName(path, name)))
+      .to(Try(configuration.get[String](path).toBoolean).toOption.getOrElse(configException(path))) //We need to parse as string, due to the process of adding in from app-config-<env> it is seen as a string
```
PeterPerhac: I think we need to do `configuration.get[Boolean](path)` and that will naturally explode with relevant ConfigurationException when the configuration isn't found under that path, or fails to be read (bad data type, etc.)

DanielConnelly: we need to get the string due to the nature of app-config-<env> passing the values.

PeterPerhac: configuration object implements `get` like so

```scala
  def get[A](path: String)(implicit loader: ConfigLoader[A]): A = {
    loader.load(underlying, path)
  }
```
so if you say get[Boolean] it defers the String->Boolean conversion to the ConfigLoader[Boolean]. I am not sure I follow, but surely we can configure boolean flags as non-strings?

app-config-... are yaml files, no? and YAML, for laughs, supports 12 different values for Boolean: https://yaml.org/type/bool.html

DanielConnelly: will discuss monday

reddy-yattapu:
:+1:

PR #742, opened 2019-11-22
Vtcca 2980

reddy-yattapu:


conf/application-json-logger.xml
```diff
@@ -15,7 +15,7 @@
 
     <logger name="uk.gov" level="${logger.uk.gov:-WARN}"/>
 
-    <root level="${logger.application:-WARN}">
+    <root level="${logger.application:-DEBUG}">
```
DanielConnelly: lets not enable DEBUG as default please.

conf/application.conf
```diff
@@ -74,6 +74,7 @@ featureFlags {
   editNameEnabled = true
   fileUploadEnabled = false
   dvrEnabled = true
+  externalDetailedValuationApiEnabled = true
```
DanielConnelly: was this work not done as part of my PR?

app/services/BusinessRatesAttachmentsService.scala
```diff
@@ -72,13 +73,48 @@ class BusinessRatesAttachmentsService @Inject()(
...
+          UploadingScanResults,
+          UploadScanResultsFailed,
+          UploadScanResultsComplete
+        ).contains(attachment._2.state))
+        Either.cond(uploaded.isEmpty, attachments, SomeFilesAreAlreadyUploaded(notuploaded.map(_._1)))
+      }
       .map(result => result.filter(_._2.state == MetadataPending))
       .subflatMap(filtered => Either.cond(filtered.size == nonEmptyReferences.size, filtered.map(_._1), NotAllFilesReadyToUpload))
       .semiflatMap(references => Future.traverse(references)(patchMetadata(submissionId, _)))
+      .leftFlatMap{
```
PeterPerhac: you can move the if-condition to the left of the arrow and then you don't need the else clause and the repeated leftT(error)

```scala
.leftFlatMap{
        case AllFilesAreAlreadyUploaded(attachments)  => EitherT.rightT(attachments)
        case error @SomeFilesAreAlreadyUploaded(references)  if (retryCount < 5) =>  submit(submissionId, references, retryCount + 1)
        case error                                    =>    EitherT.leftT(error)
      }
```

PeterPerhac:
looks good to me, just need to resolve conflict with dan's work  and if want you can do that small tweak in leftFlatMap like I suggested... 👍

app/services/BusinessRatesAttachmentsService.scala
```diff
@@ -72,13 +73,48 @@ class BusinessRatesAttachmentsService @Inject()(
...
-              nonEmptyReferences: List[String]
+              nonEmptyReferences: List[String],
+              retryCount: Int = 0
             )(implicit request: LinkingSessionRequest[_], hc: HeaderCarrier): EitherT[Future, AttachmentException, List[Attachment]] = {
     EitherT.fromEither[Future](Either.cond(nonEmptyReferences.nonEmpty, nonEmptyReferences, MissingRequiredNumberOfFiles))
       .semiflatMap(Future.traverse(_)(r => getAttachment(r).map(r -> _)))
+      .subflatMap(attachments => {
+        Either.cond(
+          attachments.filter(
+            attachment =>
```
PeterPerhac: this chunk is repeated twice and so if the states change this will rely on someone remembering to update the list in two different places. , also the contains check is performed twice. you could have a function Attachment => Boolean that has that List of states in it and checks for the contains, then just use the function in these two places

PR #743, opened 2019-11-25
VTCCA-2980 Test case fixed

reddy-yattapu:


test/controllers/ViewAssessmentSpec.scala
```diff
@@ -43,7 +43,8 @@ class ViewAssessmentSpec extends VoaPropertyLinkingSpec with OptionValues {
     mockDvrCaseManagement,
     StubBusinessRatesValuation,
     mockBusinessRatesAuthorisation,
-    stubMessagesControllerComponents()) {
+    stubMessagesControllerComponents(),
+    false) {
```
PeterPerhac: somehow this false fixes something, but it's not clear how. you could use named parameters, especially for boolean parameters

habeeb01:
👍

test/controllers/ViewAssessmentSpec.scala
reddy-yattapu: I have added name to Flag is isExternalValuationEnabled

PeterPerhac:
👍  happy

PR #744, opened 2019-11-25
Vtcca 3119

reddy-yattapu:


app/services/BusinessRatesAttachmentsService.scala
```diff
@@ -72,13 +73,48 @@ class BusinessRatesAttachmentsService @Inject()(
...
+              nonEmptyReferences: List[String],
+              retryCount: Int = 0
             )(implicit request: LinkingSessionRequest[_], hc: HeaderCarrier): EitherT[Future, AttachmentException, List[Attachment]] = {
     EitherT.fromEither[Future](Either.cond(nonEmptyReferences.nonEmpty, nonEmptyReferences, MissingRequiredNumberOfFiles))
       .semiflatMap(Future.traverse(_)(r => getAttachment(r).map(r -> _)))
+      .subflatMap(attachments => {
+        Either.cond(
+          attachments.filter(
+            attachment =>
+              List(
```
PeterPerhac: this chunk is repeated twice and so if the states change this will rely on someone remembering to update the list in two different places. , also the `contains` check is performed twice. you could have a function `Attachment => Boolean` that has that List of states in it and checks for the `contains`, then just use the function in these two places

app/services/BusinessRatesAttachmentsService.scala
```diff
@@ -72,13 +73,48 @@ class BusinessRatesAttachmentsService @Inject()(
...
+          UploadScanResultsComplete
+        ).contains(attachment._2.state))
+        Either.cond(uploaded.isEmpty, attachments, SomeFilesAreAlreadyUploaded(notuploaded.map(_._1)))
+      }
       .map(result => result.filter(_._2.state == MetadataPending))
       .subflatMap(filtered => Either.cond(filtered.size == nonEmptyReferences.size, filtered.map(_._1), NotAllFilesReadyToUpload))
       .semiflatMap(references => Future.traverse(references)(patchMetadata(submissionId, _)))
+      .leftFlatMap{
+        case AllFilesAreAlreadyUploaded(attachments)  => EitherT.rightT(attachments)
+        case error @SomeFilesAreAlreadyUploaded(references)  => if (retryCount < 5) submit(submissionId, references, retryCount + 1) else EitherT.leftT(error)
```
PeterPerhac: can move the if-condition to the left of the arrow to get rid of the else clause

conf/application.conf
```diff
@@ -86,7 +86,7 @@ enrolment {
 
 
 feature {
-  externalValuation = false
+  externalValuation = true
```
DanielConnelly: this should be false by default.

app/config/ApplicationConfig.scala
```diff
@@ -71,6 +71,7 @@ class ApplicationConfig @Inject()(configuration: Configuration, runMode: RunMode
   lazy val fileUploadEnabled: Boolean = loadBooleanConfig("featureFlags.fileUploadEnabled")
   lazy val downtimePageEnabled: Boolean = loadBooleanConfig("featureFlags.downtimePageEnabled")
   lazy val dvrEnabled: Boolean = loadBooleanConfig("featureFlags.dvrEnabled")
+  lazy val externalDetailedValuationApiEnabled: Boolean = loadBooleanConfig("featureFlags.externalDetailedValuationApiEnabled")
```
DanielConnelly: this is not used anywhere?

DanielConnelly:
LGTM 👍

PR #745, opened 2019-11-27
VTCCA-3130 matching modernised contract for the organisation

DanielConnelly:


darrell-plant:
LGTM 👍

app/controllers/agentAppointment/AppointAgentController.scala
```diff
@@ -255,9 +256,9 @@ class AppointAgentController @Inject()(
                 .map(oar => oar.copy(filterTotal = oar.authorisations.size))
                 .map(oar => oar.copy(authorisations = oar.authorisations.take(pagination.pageSize)))
             } yield {
-              BadRequest(views.html.propertyrepresentation.revokeAgentProperties(Some(errors), AppointAgentPropertiesVM(group, response), PaginationParameters(), GetPropertyLinksParameters(), group.agentCode))
+              BadRequest(views.html.propertyrepresentation.revokeAgentProperties(Some(errors), AppointAgentPropertiesVM(group, response), PaginationParameters(), GetPropertyLinksParameters(), agentCode))
             }
-          case None =>
+          case None | Some(_) =>
```
PeterPerhac: ` case None | Some(_) ` ?  if it either is or isn't there?  use `case _` for catch-all

app/controllers/agentAppointment/AppointAgentController.scala
```diff
@@ -198,9 +199,9 @@ class AppointAgentController @Inject()(
       },
       success = (agent: AgentId) => {
         accounts.withAgentCode(agent.id) flatMap {
-          case Some(group) =>
-            Future.successful(Redirect(routes.AppointAgentController.selectAgentPropertiesSearchSort(PaginationParameters(), GetPropertyLinksParameters(), group.agentCode)))
-          case None =>
+          case Some(AgentGroupAccount(_, agentCode)) =>
+            Future.successful(Redirect(routes.AppointAgentController.selectAgentPropertiesSearchSort(PaginationParameters(), GetPropertyLinksParameters(), agentCode)))
+          case None | Some(_) =>
```
habeeb01: case _ for catch all

app/models/GroupAccount.scala
```diff
@@ -18,9 +18,25 @@ package models
...
+                         isAgent: Boolean,
+                         agentCode: Option[Long]
+                       )
 
 object GroupAccount {
   implicit val format = Json.format[GroupAccount]
+
+  object AgentGroupAccount {
+    def unapply(account: GroupAccount): Option[(GroupAccount, Long)] =
+      account
```
PeterPerhac: `PartialFunction.condOpt(account){ case a@GroupAccount(_, _, _, _, _, _, true, code) => (a, code)}` avoids mapping the agent code to a tuple if the account is not actually an agent account. but i guess it's less readable

PeterPerhac:
some small nitpicks but otherwise looks good to me 👍

PR #746, opened 2019-11-28
VTCCA-3121: add test support endpoint for attachments

PeterPerhac:


darrell-plant:
No tests required? If not, then lgtm 👍

PeterPerhac:
it's a test endpoint. no other test endpoints have tests. these are our internal test endpoints. not like the ones that get exposed to 3rd parties for development in Sandbox.

PR #753, opened 2020-01-10
VTCCA-3135 Valuation Summary move

darrell-plant:


app/controllers/Assessments.scala
```diff
@@ -95,12 +96,16 @@ class Assessments @Inject()(
         logger.warn("property link assessment call failed", e)
         val linkF = if (owner) propertyLinks.getMyOrganisationPropertyLink(submissionId) else propertyLinks.getMyClientsPropertyLink(submissionId)
         linkF.map {
-          case Some(link) => Redirect(routes.Assessments.viewSummary(link.uarn, true))
+          case Some(link) => Redirect(getViewSummaryCall(link.uarn, pending = true, owner))
```
DanielConnelly: is there a reason we are not using reverse router?

DanielConnelly:
LGTM 👍

app/controllers/Assessments.scala
darrell-plant: we are using the reverse router

DanielConnelly: coolio.

reddy-yattapu:
:+1:

PR #754, opened 2020-01-27
VTCCA-0 adding an error message when not all files are ready to upload

DanielConnelly:


test/controllers/VoaPropertyLinkingSpec.scala
```diff
@@ -121,13 +121,14 @@ trait VoaPropertyLinkingSpec
...
-  override protected def beforeEach(): Unit = {
-    StubIndividualAccountConnector.reset()
-    StubGroupAccountConnector.reset()
-    StubIdentityVerification.reset()
-    StubPropertyLinkConnector.reset()
-    StubBusinessRatesValuation.reset()
-    StubSubmissionIdConnector.reset()
-    StubPropertyRepresentationConnector.reset()
-  }
+//  override protected def beforeEach(): Unit = {
```
darrell-plant: Needs a tidy up?

PeterPerhac: commented out code

DanielConnelly: updated.

DanielConnelly: updated.

darrell-plant:
LGTM 👍

conf/messages
```diff
@@ -707,6 +707,7 @@ declaration.p.3=Once you submit the form you will not be able to go back and mak
 label.declaration=Declaration
 declaration.iAgree=I agree to the above declaration
 declaration.required=You must agree to the declaration to continue
+declaration.file.receipt=Awaiting evidence receipt, please try again in a moment.
```
PeterPerhac: I don't think this makes it any clearer for the user... I wouldn't know what that means. I tick a box to declare something, press submit, and back to the same page with "Awaiting evidence receipt ..." I'd be well confused. Has the text come from Karen?

darrell-plant: It is clearer. But could be worded better.

PR #768, opened 2020-03-04
Vtcca 3209

habeeb01:


app/connectors/propertyLinking/PropertyLinkConnector.scala
```diff
@@ -177,6 +177,15 @@ class PropertyLinkConnector @Inject()(config: ServicesConfig, http: HttpClient)(
   def getMyOrganisationAgents()(implicit hc: HeaderCarrier): Future[AgentList] =
     http.GET[AgentList](s"$baseUrl/owner/agents")
 
+  def sendAgentRelationshipRequest(agentRelationshipRequest: AgentRelationshipRequest)(
+        implicit hc: HeaderCarrier): Future[Unit] = {
+    http
+      .POST[AgentRelationshipRequest, HttpResponse](
+        s"$baseUrl/my-organisation/appointmentChanges",
+        agentRelationshipRequest)
+      .map(_ => ()) //fixme is it async ?
```
DanielConnelly: this is async. just capture the response (which will be a generic success response) and pass this up.

app/controllers/agentAppointment/AgentRelationshipController.scala
```diff
@@ -78,10 +78,10 @@ class AgentRelationshipController @Inject()(
       },
       success => {
```
DanielConnelly: why is this not already long?

app/controllers/agentAppointment/AgentRelationshipController.scala
```diff
@@ -78,10 +78,10 @@ class AgentRelationshipController @Inject()(
...
         for {
-          organisationsAgents <- agentRelationshipService.getMyOrganisationAgents()
-          agentOpt            <- agentRelationshipService.getAgent(success.toLong)
+          organisationsAgents    <- agentRelationshipService.getMyOrganisationAgents()
```
DanielConnelly: why are you getting this?

conf/app.routes
```diff
@@ -106,6 +106,7 @@ GET         /my-organisation/appoint-new-agent/multiple-properties
 POST        /my-organisation/appoint-new-agent/multiple-properties                                                        @controllers.agentAppointment.AgentRelationshipController.submitMultipleProperties()
 
 GET         /my-organisation/appoint-new-agent/checkYourAnswers                                                           @controllers.agentAppointment.AgentRelationshipController.checkAnswers()
+POST        /my-organisation/appoint-new-agent/appointmentChanges                                                         @controllers.agentAppointment.AgentRelationshipController.submitAppointmentChanges()
```
DanielConnelly: as this method is used for revoke as well, we cannot call this appoint.

app/connectors/propertyLinking/PropertyLinkConnector.scala
habeeb01: addressed

app/controllers/agentAppointment/AgentRelationshipController.scala
habeeb01: im getting it because I need to determine if the agent has already been added to the org

conf/app.routes
habeeb01: good point

app/connectors/propertyLinking/PropertyLinkConnector.scala
DanielConnelly: I also dont know if we should be expanding there pattern all the way to the frontend. if we had several endpoints in voa-property-linking backend that do appoint and revoke seperately it might make us less coupled to modernised behaviour.

app/controllers/agentAppointment/AgentRelationshipController.scala
```diff
@@ -206,12 +213,41 @@ class AgentRelationshipController @Inject()(
   }
 
   def checkAnswers(): Action[AnyContent] = authenticated.andThen(withAppointAgentSession).async { implicit request =>
```
DanielConnelly: dont use .async here and you dont have to wrap in future.

app/controllers/agentAppointment/AgentRelationshipController.scala
```diff
@@ -206,12 +213,41 @@ class AgentRelationshipController @Inject()(
...
-    Future.successful(Ok(checkYourAnswers(request.sessionData)))
+    Future.successful(Ok(checkYourAnswers(submitAgentRelationship, request.sessionData)))
+  }
+
+  def submitAppointmentChanges(): Action[AnyContent] = authenticated.andThen(withAppointAgentSession).async {
```
DanielConnelly: are we having only one endpoint in the frontend that does both? I am not 100% sure. that leaves us protected and makes the testng complicated. would it be possible to split this up making our testing jobs easier.

DanielConnelly: golden rule, one endpoint one function.

app/controllers/agentAppointment/AgentRelationshipController.scala
```diff
@@ -206,12 +213,41 @@ class AgentRelationshipController @Inject()(
...
+        }, { success =>
+          {
+            agentRelationshipService.sendAgentRelationshipRequest(
+              AgentRelationshipRequest(
+                agentRepresentativeCode = success.agentCode,
+                action = AppointmentAction.withName(success.action),
+                scope = AppointmentScope.withName(success.scope),
+                propertyLinks = None
+              )
+            )
```
DanielConnelly: why is this not in a .map.

app/controllers/agentAppointment/AgentRelationshipController.scala
```diff
@@ -206,12 +213,41 @@ class AgentRelationshipController @Inject()(
...
+            Future.successful(Ok(confirmation(getAgentName(request))))
+          }
+        }
+      )
   }
 
   private def getAgentName(request: AppointAgentSessionRequest[AnyContent]) =
```
DanielConnelly: this should be a generic function get Agent. which returns the agent details object.

DanielConnelly: And if we require the agent address here, add a seperate refiner step on the methods that call this.

app/models/propertyrepresentation/AppointmentScope.scala
```diff
@@ -0,0 +1,30 @@
...
+
+package models.propertyrepresentation
+
+import play.api.libs.json.Format
+import utils.JsonUtils
+
+object AppointmentScope extends Enumeration {
+  type AppointmentScope = Value
+
+  val RELATIONSHIP = Value("RELATIONSHIP")
```
DanielConnelly: let discuss this tomorrow.

app/views/propertyrepresentation/appoint/isThisYourAgent.scala.html
```diff
@@ -22,7 +22,7 @@
 
 @this(govukErrorSummary : GovukErrorSummary, govukRadios : GovukRadios, govukButton : GovukButton, mainLayout: views.html.mainLayout)
 
-@(form: Form[Boolean], agentName: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
+@(form: Form[Boolean], agentName: String, agentAddress: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
```
DanielConnelly: should pass the agent details into page. cleans up parameters.

app/controllers/agentAppointment/AppointNewAgentForms.scala
```diff
@@ -39,4 +39,12 @@ object AppointNewAgentForms {
   val manageMultipleProperties: Form[ManagePropertiesOptions] = Form(
     Forms.single("multipleProperties" -> EnumMapping(ManagePropertiesOptions, "error.multipleProperties.required")))
 
+  val submitAgentRelationship: Form[AgentRelationshipForm] =
```
DanielConnelly: I dont see any test cases for this form.

app/controllers/agentAppointment/AgentRelationshipController.scala
habeeb01: i meant to change this. It was getAgentName when we only had the name. I started using Jon's agent details endpoint yesterday. Meant to refactor this.

app/views/propertyrepresentation/appoint/isThisYourAgent.scala.html
habeeb01: agreed

app/controllers/agentAppointment/AppointNewAgentForms.scala
habeeb01: there are tests in AgentRelationshipControllerSpec

app/controllers/agentAppointment/AgentRelationshipController.scala
habeeb01: it does only do one thing, as per design 😄

habeeb01: i do agree that we could do it differently. I will create separate methods for appoint and revoke

app/controllers/agentAppointment/AgentRelationshipController.scala
habeeb01: wrapped in Try

app/actions/agentrelationship/request/AppointAgentSessionRequest.scala
```diff
@@ -41,4 +41,9 @@ case class AppointAgentSessionRequest[A](
   }
 
   def organisationId = groupAccount.id
+
+  def agentDetails = AgentDetails(
+    name = sessionData.agentOrganisationName.getOrElse(throw new IllegalStateException("no agent name stored")),
+    address = sessionData.agentAddress.getOrElse(throw new IllegalStateException("no agent address stored"))
+  )
```
darrell-plant: Feels like we should have a CcaAgentWrappedRequest or something to avoid these exceptions?

app/models/propertyrepresentation/AgentDetails.scala
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package models.propertyrepresentation
+
+import play.api.libs.json.{Json, OFormat}
+
+case class AgentDetails(name: String, address: String)
+
+object AgentDetails {
+  implicit val format: OFormat[AgentDetails] = Json.format[AgentDetails]
```
darrell-plant: implicit val format: OFormat[AgentDetails] = Json.format

app/models/propertyrepresentation/AgentRelationshipForm.scala
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package models.propertyrepresentation
+
+import play.api.libs.json.{Json, OFormat}
+
+case class AgentRelationshipForm(agentCode: Long, action: String, scope: String)
+
+object AgentRelationshipForm {
+  implicit val format: OFormat[AgentRelationshipForm] = Json.format[AgentRelationshipForm]
```
darrell-plant: implicit val format: OFormat[AgentRelationshipForm] = Json.format

app/views/propertyrepresentation/appoint/checkYourAnswers.scala.html
```diff
@@ -20,14 +20,26 @@
...
 
-@(data: AppointNewAgentSession)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
+@(form: Form[AppointAgentRequest], data: AppointNewAgentSession)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
 
 @answerMessageKey = @{
     s"propertyRepresentation.checkYourAnswers.options.${data.managingProperty.getOrElse("")}"
 }
+@appointmentScope = @{
+    data.managingProperty match {
+        case None => throw new IllegalStateException("Unable to get 'managingProperty' from session data")
```
darrell-plant: Isn't this the kind of thing we can avoid by using a request wrapper that guarantees this data is available?

test/forms/AppointAgentRequestSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package forms
+
+import models.propertyrepresentation.{AppointAgentRequest, AppointmentScope}
+import org.scalatest.{FlatSpec, MustMatchers}
+import utils.FormBindingVerification._
+import views.helpers.Errors
+
+class AppointAgentRequestSpec extends FlatSpec with MustMatchers {
```
darrell-plant: FlatSpec is not our standard.

app/models/propertyrepresentation/AgentRelationshipForm.scala
habeeb01: AgentRelationshipForm has now been deleted. I replaced it in my previous commit

test/forms/AppointAgentRequestSpec.scala
habeeb01: I am now extending VoaPropertyLinkingSpec. It still used FlatSpec 🤷‍♂

app/views/propertyrepresentation/appoint/checkYourAnswers.scala.html
habeeb01: i think we might want to blow up if the data is not there at this point in the journey as it would mean that something is wrong. May be @DanielConnelly wants to weigh in on this

darrell-plant: It would be better if we never got this far if we didn't have the required data. Feels like we need a new ActionFilter (or similar) that ensures that agent data is available? But yes, get @DanielConnelly 's opinion.

DanielConnelly: We should never knowingly allow a 500 exception. it is easier to manage these errors in what darrell suggested. a filter with a new request type would allow you to be sure only those get in correctly.

app/controllers/agentAppointment/AgentRelationshipController.scala
DanielConnelly: > it does only do one thing, as per design 😄

the design say it APPOINT's and REVOKES which are at a base level two different functions.

app/models/propertyrepresentation/AppointNewAgentSession.scala
```diff
@@ -16,15 +16,78 @@
...
-case class AppointNewAgentSession(
+sealed trait AppointNewAgentSession {
+  val status: String
+}
+
+case class SearchedAgent(
+      agentCode: Long,
+      agentOrganisationName: String,
+      agentAddress: String,
+      status: String = "SearchedAgent")
```
DanielConnelly: why are we not using Enumerations for these?

app/models/propertyrepresentation/AppointNewAgentSession.scala
```diff
@@ -16,15 +16,78 @@
...
 object AppointNewAgentSession {
-  implicit val format: OFormat[AppointNewAgentSession] = Json.format[AppointNewAgentSession]
+
+  val readers: Map[String, Reads[_ <: AppointNewAgentSession]] = Map(
+    "SearchedAgent"    -> implicitly[Reads[SearchedAgent]],
+    "SelectedAgent"    -> implicitly[Reads[SelectedAgent]],
+    "ManagingProperty" -> implicitly[Reads[ManagingProperty]]
+  )
+
+  implicit val animalReads: Reads[AppointNewAgentSession] = new Reads[AppointNewAgentSession] {
```
DanielConnelly: animal? also where is the writes.

habeeb01: copy and paste error from an example 🤦‍♂. I don't think I need the writes. Will it not use the format from the inheriting case classes ?

app/controllers/agentAppointment/AgentRelationshipController.scala
```diff
@@ -193,38 +215,72 @@ class AgentRelationshipController @Inject()(
...
   }
 
-  private def getAgentName(request: AppointAgentSessionRequest[AnyContent]) =
-    request.sessionData.agentOrganisationName.getOrElse(throw new IllegalStateException("no agent name stored"))
+  def appointAgent(): Action[AnyContent] = authenticated.andThen(withAppointAgentSession).async { implicit request =>
+    submitAppointAgentRequest.bindFromRequest.fold(
+      errors => {
+        PartialFunction
+          .condOpt(request.sessionData) {
+            case data: ManagingProperty =>
```
darrell-plant: I'm a bit confused about the PartialFunction.condOpt
Can you not just match on request.sessionData?

darrell-plant:
One comment, but LGTM 👍

app/controllers/agentAppointment/AgentRelationshipController.scala
habeeb01: IMHO, it would have the same effect. PartialFunction.condOpt was suggested by Dan.

app/models/propertyrepresentation/AppointNewAgentSession.scala
habeeb01: now using enums

app/controllers/agentAppointment/AgentRelationshipController.scala
darrell-plant: Chatted with @DanielConnelly and decided changing this back to a straight match on request.sessionData would be better.

app/models/propertyrepresentation/AppointNewAgentSession.scala
DanielConnelly: 👍

DanielConnelly:
LGTM 👍

PR #774, opened 2020-03-16
Vtcca 3204

habeeb01:


app/models/propertyrepresentation/ManageAgentSession.scala
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.propertyrepresentation
+
+//case class AppointNewAgentSession(
```
DanielConnelly: commented code.

reddy-yattapu:
:+1:

app/controllers/agent/ManageAgentController.scala
```diff
@@ -0,0 +1,131 @@
...
+      override val controllerComponents: MessagesControllerComponents,
+      executionContext: ExecutionContext,
+      val config: ApplicationConfig
+) extends PropertyLinkingController {
+
+  val logger = Logger(this.getClass.getName)
+
+  def manageAgent(agentCode: Option[Long]): Action[AnyContent] = authenticated.async { implicit request =>
+    getManageAgentPage(agentCode).map { pageOpt =>
+      pageOpt match {
```
DanielConnelly: you can do an anonymous function here .map { case None => .... case Some(page) => ... }

app/controllers/agent/ManageAgentController.scala
```diff
@@ -0,0 +1,131 @@
...
+      manageAgentOnePropertyLinkWithAgentAssigned: views.html.propertyrepresentation.manage.manageAgentOnePropertyLinkWithAgentAssigned)(
+      implicit override val messagesApi: MessagesApi,
+      override val controllerComponents: MessagesControllerComponents,
+      executionContext: ExecutionContext,
+      val config: ApplicationConfig
+) extends PropertyLinkingController {
+
+  val logger = Logger(this.getClass.getName)
+
+  def manageAgent(agentCode: Option[Long]): Action[AnyContent] = authenticated.async { implicit request =>
```
DanielConnelly: why is agent code optional?

app/controllers/agent/ManageAgentController.scala
```diff
@@ -0,0 +1,131 @@
...
+                                         searchParams = GetPropertyLinksParameters(),
+                                         pagination = PaginationParams(
+                                           startPoint = 1,
+                                           pageSize = 1000,
+                                           requestTotalRowCount = false),
+                                         representationStatusFilter =
+                                           Seq(RepresentationApproved, RepresentationPending)
+                                       )
+    } yield {
+      (propertyLinks.total, agentToBeManagedOpt) match {
```
DanielConnelly: I dont fully understand what this section is doing, could you maybe explain the logic here.

app/controllers/agent/ManageAgentForms.scala
```diff
@@ -0,0 +1,27 @@
...
+
+package controllers.agent
+
+import form.EnumMapping
+import models.propertyrepresentation.ManageAgentOptions
+import play.api.data.{Form, Forms}
+
+object ManageAgentForms {
+
+  val manageAgent: Form[ManageAgentOptions] = Form(Forms.single("manageAgent" -> EnumMapping(ManageAgentOptions)))
```
DanielConnelly: is this a form for ManageAgentForms or ManageAgentOptions? the singleton object for this form has me confused.

app/controllers/agent/ManageAgentController.scala
habeeb01: it is optional because we won't necessarily have the agentCode at the start. If you look the table in the story (https://jira.tools.tax.service.gov.uk/browse/VTCCA-3204). The scenarios where we are meant to skip the "new agents list" page.

app/models/propertyrepresentation/ManageAgentOptions.scala
```diff
@@ -0,0 +1,125 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.propertyrepresentation
+
+import models.{EnumFormat, NamedEnum, NamedEnumSupport}
+import play.api.libs.json.{Json, OFormat}
+
+sealed trait ManageAgentOptions extends NamedEnum {
```
DanielConnelly: what is the difference between these options and the above?

app/controllers/agent/ManageAgentController.scala
habeeb01: I will add comments to make it clearer. Each case is basically a scenario in the story

app/controllers/agent/ManageAgentForms.scala
habeeb01: ManageAgentForms is meant to be deleted, it is not used anymore. It is part of the refactor i did this. I forgot to remove it

habeeb01: im using ManageAgentRequest instead

app/models/propertyrepresentation/ManageAgentOptions.scala
habeeb01: answered above

app/controllers/agent/ManageAgentController.scala
DanielConnelly: the scenario where the user only has one property?

app/controllers/agent/ManageAgentController.scala
```diff
@@ -0,0 +1,137 @@
...
+              ManageAgentOptions.multiplePropertyLinksAgentAssignedToAllOptions,
+              agent))
+        case _ => None
+      }
+    }
+
+  def submitManageAgent(): Action[AnyContent] = authenticated.async { implicit request =>
+    ManageAgentRequest.submitManageAgentRequest.bindFromRequest.fold(
+      errors => {
+        val agentCodeFromForm = Try(errors.data("agentCode").toLong).toOption
```
DanielConnelly: this looks slightly dodgy, should we not be passing the agentCode in the URL that way we can be sure the data will be there.

app/controllers/agent/ManageAgentController.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  def submitManageAgent(agentCode: Long): Action[AnyContent] = authenticated.async { implicit request =>
+    ManageAgentRequest.submitManageAgentRequest.bindFromRequest.fold(
+      errors => {
+        getManageAgentPage(Some(agentCode), errors).map {
+          case None       => NotFound(errorHandler.notFoundTemplate)
+          case Some(page) => BadRequest(page)
+        }
+      }, { success =>
+        ??? //fixme this will be implemented in subsequent stories
```
darrell-plant: Could return 501 Not Implemented
Any Jira number for the story that will implement this?

app/controllers/agentAppointment/AddAgentController.scala
```diff
@@ -16,31 +16,41 @@
...
+import models.searchApi.AgentPropertiesParameters
+import views.html.propertyrepresentation.appoint._
+import controllers.agentAppointment.AppointNewAgentForms._
 import uk.gov.hmrc.propertylinking.errorhandler.CustomErrorHandler
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 
-class AgentRelationshipController @Inject()(
+class AddAgentController @Inject()(
```
darrell-plant: Looks like a lot of unnecessary {}'s in this file.

app/models/propertyrepresentation/AddAgentOptions.scala
```diff
@@ -18,34 +18,34 @@ package models.propertyrepresentation
 
 import models.{EnumFormat, NamedEnum, NamedEnumSupport}
 
-sealed trait ManagePropertiesOptions extends NamedEnum {
+sealed trait AddAgentOptions extends NamedEnum {
```
darrell-plant: Shouldn't we be using Enumeration?

app/models/propertyrepresentation/ManageAgentOptions.scala
darrell-plant: Shouldn't we be using Enumeration?

test/controllers/agent/ManageAgentControllerSpec.scala
```diff
@@ -0,0 +1,195 @@
...
+import play.api.test.{FakeRequest, Helpers}
+import play.api.test.Helpers._
+import play.twirl.api.Html
+import services.AgentRelationshipService
+import tests.AllMocks
+import uk.gov.hmrc.propertylinking.errorhandler.CustomErrorHandler
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class ManageAgentControllerSpec extends VoaPropertyLinkingSpec with MockitoSugar with AllMocks {
```
darrell-plant: This doesn't look like most of our tests. It doesn't have the same layout/structure. No shared setups. No `when`s

test/controllers/agentAppointment/AddAgentControllerSpec.scala
```diff
@@ -32,7 +32,7 @@ import play.twirl.api.Html
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
-class AgentRelationshipControllerSpec extends VoaPropertyLinkingSpec with MockitoSugar with AllMocks {
+class AddAgentControllerSpec extends VoaPropertyLinkingSpec with MockitoSugar with AllMocks {
```
darrell-plant: Same as above

test/utils/FakeObjects.scala
```diff
@@ -204,10 +204,10 @@ trait FakeObjects {
...
   val ownerAuthResultWithNoAuthorisations = OwnerAuthResult(
-    start = 1,
-    size = 10,
-    filterTotal = 10,
-    total = 10,
+    start = 0,
+    size = 0,
+    filterTotal = 0,
+    total = 0,
     authorisations = Seq.empty
```
darrell-plant: start and size = 0 are not realistic (allowed) values are they?

test/controllers/agent/ManageAgentControllerSpec.scala
habeeb01: It does. please have a look at other tests in voa-property-linking-frontend

test/controllers/agentAppointment/AddAgentControllerSpec.scala
habeeb01: same as above

app/controllers/agentAppointment/AddAgentController.scala
habeeb01: I have added braces to help with readability. I think they make this clearer in this case. Which ones particularly bother you?

app/models/propertyrepresentation/AddAgentOptions.scala
habeeb01: NamedEnum seem to be the standard in voa-property-linking-frontend

test/controllers/agent/ManageAgentControllerSpec.scala
darrell-plant: I was referring to new standards, not old ones :)

test/controllers/agentAppointment/AddAgentControllerSpec.scala
darrell-plant: see above :)

app/controllers/agentAppointment/AddAgentController.scala
darrell-plant: You may have noticed that braces often get removed other devs? My impression is that the common standard on this team is to avoid unnecessary {}'s. You added braces where they weren't before. The next dev may remove them. Where does it end :) Not a big deal obviously, and none of them "particularly bother" me.

test/utils/FakeObjects.scala
habeeb01: can you give me an example of realistic values for when you are returning no authorisations?

app/models/propertyrepresentation/AddAgentOptions.scala
darrell-plant: Sometimes things need to change.

test/utils/FakeObjects.scala
darrell-plant: Yes, start = 1 and size = 15.

habeeb01: how can start and size have those values when there are no auths? what do those numbers relate to ?

darrell-plant: Do you not know? You changed the values so I assumed you knew. start is the page number (min value 1), size is the page size (min value 10, default 15)

app/models/propertyrepresentation/AddAgentOptions.scala
habeeb01: im happy to change it if you can let me know what we will gain from making that change. If its just a case of preference, then I will keep it as it is

app/controllers/agentAppointment/AddAgentController.scala
PeterPerhac: these are unnecessary braces that even intellij highlights to you with a hint to "Remove redundant braces". As per official Scala 
Style Guide https://docs.scala-lang.org/style/control-structures.html#curlybraces 
> case - Always omit braces in case clauses.

image

app/controllers/agentAppointment/AddAgentController.scala
```diff
@@ -76,20 +86,21 @@ class AgentRelationshipController @Inject()(
                     data = Map("agentCode" -> success),
                     errors = Seq[FormError](FormError(key = "agentCode", message = "error.agentCode.required")))
                 )))
-          case Some(representativeCode) =>
+          case Some(representativeCode) => {
```
PeterPerhac: Always omit braces in case clauses.

app/models/propertyrepresentation/AddAgentOptions.scala
darrell-plant: It's about standards and patterns. I'm confused when I see different approaches to doing the same thing. 
Here I am asking a question "Shouldn't we be using Enumeration?" - not telling you to change it. Also, this has been raised before in PR and I seem to remember Dan saying we should use Enumeration?

darrell-plant: Maybe some of the discussion points here could be part of Dan's handover?

test/utils/FakeObjects.scala
habeeb01: i changed it to what I thought they should be. Are those values meant to be copied from the request? I don't see why we would set them to anything other than zero. Do those numbers not relate to the face that we are returning no results?

darrell-plant: I'm not sure I can explain any better than this:`start` is the page number (min value 1), `size` is the page size (min value 10, default 15)

darrell-plant: they are pagination parameters aren't they? which page you are on, and max results per page

habeeb01: ive changed it to what you suggested (it still doesn't make sense to me but 🤷‍♂)

DanielConnelly: @darrell-plant and @habeeb01 if you guys can have a quick call around this to clear up the confusion.

darrell-plant: When we make a request to a paginated endpoint , we supply the page number and page size. Page size is restricted to be between 10 and 100 (I think) and usually defaults to 15. page 0 doesn't exist :)
The response returns the page and size values that were supplied so that the client code can know where they are in the pagination and don't have to store previous values.

A response with no results will always be on page 1 and the requested page size would be 10 or 15 or ...

darrell-plant:
Apart from various discussion points that we can pick up in the handover this LGTM 👍

PR #777, opened 2020-03-19
VTCCA-3207: add existing agent to all properties

habeeb01:


app/views/propertyrepresentation/manage/addAgentToAllProperty.scala.html
```diff
@@ -0,0 +1,53 @@
...
+@this(govukErrorSummary : GovukErrorSummary, govukRadios : GovukRadios, govukButton : GovukButton, mainLayout: views.html.mainLayout)
+
+
+@(form: Form[AppointAgentRequest], agentName: String, agentCode: Long)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
+
+
+
+@mainLayout(
+    pageTitle = messages("propertyRepresentation.addAnAgent.title"),
+    backUri = Some("#")
```
darrell-plant: Is this a TODO?

reddy-yattapu:
:+1:

conf/messages
```diff
@@ -444,7 +444,7 @@ propertyRepresentation.checkYourAnswers.submit=Submit
 propertyRepresentation.confirmation={0} has been appointed for you
 propertyRepresentation.confirmation.emailSent=We have sent you a confirmation email.
 propertyRepresentation.confirmation.whatHappensNext=What happens next
-propertyRepresentation.confirmation.whatHappensNext.p1=We will notify Knight Frank that you have appointed them as your agent.
+propertyRepresentation.confirmation.whatHappensNext.p1=We will notify {0}} that you have appointed them as your agent.
```
darrell-plant: extra }

conf/messages
```diff
@@ -464,6 +464,11 @@ propertyRepresentation.manageAgent.options.unassignFromSomeProperties=Unassign f
 propertyRepresentation.manageAgent.options.assignToYourProperty=Assign to your property
 propertyRepresentation.manageAgent.options.removeFromYourAccount=Remove from your account
 
+propertyRepresentation.manageAgent.assignToAllProperties.question=Are you sure you want to add {0} to all your properties?
```
darrell-plant: should `add` be `assign` ?

test/utils/FakeObjects.scala
```diff
@@ -203,6 +219,14 @@ trait FakeObjects {
     authorisations = Seq(ownerAuthorisation, ownerAuthorisation2)
   )
 
+  val ownerAuthResultWithTwoAuthsAgentAssignedToOne = OwnerAuthResult(
+    start = 1,
+    size = 2,
```
darrell-plant: size 15 :)

app/models/propertyrepresentation/ManageAgentRequest.scala
```diff
@@ -21,7 +21,7 @@ import play.api.data.{Form, Forms}
 import play.api.data.Forms.{longNumber, mapping, optional, text}
 import play.api.libs.json.{Json, OFormat}
 
-case class ManageAgentRequest(agentCode: Long, manageAgentOption: ManageAgentOptions)
+case class ManageAgentRequest(agentName: String, manageAgentOption: ManageAgentOptions)
```
PeterPerhac: I am surprised by this change, are we managing agents by their "name"? but there could be duplicates, no? Couldn't one agent have multiple codes? Then which agent are we managing if we try to identify them by code? Can't they update their Name?

conf/messages
PeterPerhac: double closing brace `}}`

app/views/propertyrepresentation/manage/addAgentToAllProperty.scala.html
habeeb01: good catch. No it is not.

darrell-plant:
a couple of files are badly named e.g. addAgentToAllProperty.scala.html  should probably be addAgentToAllProperties.scala.html - but not a big deal

conf/messages
habeeb01: I thought the same thing. The text came from the prototype so I don't want to change it without being told to. I will check with aunty Karen.

app/views/propertyrepresentation/manage/addAgentToAllProperty.scala.html
PeterPerhac: so a BACK link would reload the same page? If there isn't a back link, then shouldn't we use `None`? I am asking because I haven't done much frontend-y work. But I don't understand why `#` would be a back link

app/models/propertyrepresentation/ManageAgentRequest.scala
habeeb01: we are not managing by name, agentCode is what we pass to the backend. I am passing the name in the url because I need to display it on the page. I didn't want to have to go and fetch it every time (I can do that instead if you guys feel its better to do so). 

I don't think its possible for one agent to have multiple codes.

conf/messages
darrell-plant: rule number 1: never trust the prototype :)

darrell-plant:
LGTM 👍

conf/messages
habeeb01: waiting on a response from Karen on this one

PR #778, opened 2020-03-19
VTCCA-3207: appoint agent to some properties

habeeb01:


app/controllers/agent/ManageAgentController.scala
```diff
@@ -150,4 +151,16 @@ class ManageAgentController @Inject()(
...
 
+  private def joinOldJourney(agentCode: Long) =
+    Redirect(
+      controllers.agentAppointment.routes.AppointAgentController.getMyOrganisationPropertyLinksWithAgentFiltering(
+        pagination = PaginationParameters(page = 1, pageSize = 15),
+        params = GetPropertyLinksParameters(
+          sortfield = ExternalPropertyLinkManagementSortField.ADDRESS,
+          sortorder = ExternalPropertyLinkManagementSortOrder.ASC),
+        agentCode = agentCode,
+        agentAppointed = Some(Both.name),
```
PeterPerhac: what is agentAppointed = Both?

PeterPerhac:
👍 Looks good to me. Only had one question

darrell-plant:
LGTM 👍 (apart from Pete's question)

app/controllers/agent/ManageAgentController.scala
habeeb01: passing "agentAppointed = Both" as a param basically means that you are requesting all property links (from that org) regardless of whether they have an agent appointed

PR #780, opened 2020-03-20
VTCCA-3206: implemented unassign agent from some properties

habeeb01:


app/controllers/agent/ManageAgentController.scala
```diff
@@ -140,13 +140,13 @@ class ManageAgentController @Inject()(
...
         success.manageAgentOption match {
-          case AssignToSomeProperties => Future.successful(joinOldJourney(agentCode))
+          case AssignToSomeProperties => Future.successful(joinOldAgentAppointJourney(agentCode))
           case AssignToAllProperties =>
             Future.successful(Ok(addAgentToAllProperties(submitAgentAppointmentRequest, success.agentName, agentCode)))
           case UnassignFromAllProperties =>
             Future.successful(
               Ok(unassignAgentFromAllProperties(submitAgentAppointmentRequest, success.agentName, agentCode)))
-          case _ => ??? //fixme this will be implemented in subsequent stories (VTCCA-3206)
+          case UnassignFromSomeProperties => Future.successful(joinOldRevokeAppointJourney(agentCode))
```
DanielConnelly: join?

app/controllers/agent/ManageAgentController.scala
```diff
@@ -190,4 +190,15 @@ class ManageAgentController @Inject()(
         backLink = controllers.agent.routes.ManageAgentController.manageAgent(Some(agentCode)).url
       ))
 
+  private def joinOldRevokeAppointJourney(agentCode: Long) =
+    Redirect(
+      controllers.agentAppointment.routes.AppointAgentController.selectAgentPropertiesSearchSort(
```
DanielConnelly: so you are not making any changes to that old screen?

DanielConnelly: how will the choices in that screen be reflected on the new endpoint?

app/controllers/agent/ManageAgentController.scala
habeeb01: Karen said we would just join the old journey for the "some" options

app/controllers/agent/ManageAgentController.scala
habeeb01: Karen said we would just join the old journey for the "some" options. These. will still go to the old endpoint. We need to confirm with Karen if that is the case.

app/controllers/agent/ManageAgentController.scala
```diff
@@ -178,7 +179,7 @@ class ManageAgentController @Inject()(
       )
   }
 
-  private def joinOldJourney(agentCode: Long) =
+  private def joinOldAgentAppointJourney(agentCode: Long) =
     Redirect(
       controllers.agentAppointment.routes.AppointAgentController.getMyOrganisationPropertyLinksWithAgentFiltering(
         pagination = PaginationParameters(page = 1, pageSize = 15),
```
darrell-plant: PaginationParameters() defaults to PaginationParameters(page = 1, pageSize = 15)

app/controllers/agent/ManageAgentController.scala
```diff
@@ -190,4 +191,14 @@ class ManageAgentController @Inject()(
         backLink = controllers.agent.routes.ManageAgentController.manageAgent(Some(agentCode)).url
       ))
 
+  private def joinOldRevokeAppointJourney(agentCode: Long) =
+    Redirect(
+      controllers.agentAppointment.routes.AppointAgentController.selectAgentPropertiesSearchSort(
+        pagination = PaginationParameters(page = 1, pageSize = 15),
```
darrell-plant: same

app/controllers/agent/ManageAgentController.scala
PeterPerhac: is this exhaustively matched? We no longer need a catch-all clause `case _ => ` right? That's good, I am just checking...

PeterPerhac:
👍 LGTM

app/views/propertyrepresentation/appoint/confirmation.scala.html
```diff
@@ -36,7 +36,7 @@
     <p class="govuk-body">@messages("propertyRepresentation.confirmation.emailSent")</p>
 
     <h2 id="whatHappensNext" class="govuk-heading-m">@messages("propertyRepresentation.confirmation.whatHappensNext")</h2>
-    <p id="whatHappensNextP1" class="govuk-body">@messages("propertyRepresentation.confirmation.whatHappensNext.p1, agentName")</p>
+    <p id="whatHappensNextP1" class="govuk-body">@messages("propertyRepresentation.confirmation.whatHappensNext.p1", agentName)</p>
```
darrell-plant: This naming convention is not good (adding P1 and such like). Ids should tell you about the thing, not where it is currently or what html tag it currently uses. We do it all over the place, especially in the messages file, but it's not a good idea.

darrell-plant:
minor comments, but LGTM 👍

app/views/propertyrepresentation/appoint/confirmation.scala.html
PeterPerhac: 🦅 👁 Dazza. That's a good catch.

PeterPerhac: also, i don't think individual paragraphs (p tags) need an id. I thought only inputs or buttons or possibly divs might need it. only things the tests might need to be able to directly target.

habeeb01: I agree with you @darrell-plant. It ended up with that name because I couldn't come up with a better name and I wanted to move on.

Redarding @peteharpersc's point. Im only adding the id's to aid testing so I will take a steer from Colin/Vijay on what elements they want to have ids.

habeeb01: the confirmation page is different for each of the journeys in the new agent relationship stuff. So I would imagine that the QAs would want to confirm the text on the screen

habeeb01: may be we can have a general chat with the QAs re what elements should get ids

PR #784, opened 2020-03-24
VTCCA-3349: Redirect user to login if file upload initiate fails

PeterPerhac:
redirect user to sign in page when session expires on the upload rates bill page


darrell-plant:
Quick scan looks good 👍

habeeb01:
LGTM 👍

PR #787, opened 2020-03-25
VTCCA-0 resolving qa issues due to broke endpoint

DanielConnelly:


darrell-plant:
LGTM 👍

app/controllers/agentAppointment/AppointAgentController.scala
```diff
@@ -75,11 +75,10 @@ class AppointAgentController @Inject()(
       .bindFromRequest()
       .fold(
         hasErrors = errors => {
-          agentsConnector.ownerAgents(request.organisationId) map { ownerAgents =>
+          getAgents(request.organisationId) map { ownerAgents =>
```
PeterPerhac: i think we have agreed to not use infix notation for `map` `flatMap` etc. in production code. It's ok in tests to say `foo shouldBe bar` and not `foo.shouldBe(bar)` but in code we'd like to always consistently use the dot `foo.map(_ => true)`

app/controllers/agentAppointment/AppointAgentController.scala
```diff
@@ -94,11 +93,11 @@ class AppointAgentController @Inject()(
                     None)))
             case None =>
               val errors: List[FormError] = List(invalidAgentCode)
-              agentsConnector.ownerAgents(request.organisationId) flatMap { ownerAgents =>
+              getAgents(request.organisationId) flatMap { ownerAgents =>
```
PeterPerhac: i think we have agreed to not use infix notation for `map` `flatMap` etc. in production code. It's ok in tests to say `foo shouldBe bar` and not `foo.shouldBe(bar)` but in code we'd like to always consistently use the dot `foo.map(_ => true)`

PeterPerhac:
Cool, Thanks Dan! 👍 I know my comments were on someone else's original code, but I now noticed. It's up to you if you want to address the minor nitpicks.

PR #788, opened 2020-03-26
VTCCA-3349: change continueURL for linking

PeterPerhac:
when user lost their session during property linking, redirect to sign in page, with a continue URL back to start of property linking journey

### DIFF
--- a/app/controllers/IdentityVerification.scala
+++ b/app/controllers/IdentityVerification.scala
@@ -76,15 +76,19 @@ class IdentityVerification @Inject()(
   }
 
   def success(journeyId: Option[String]): Action[AnyContent] = ggAction.async { implicit request =>
-    journeyId.fold(Future.successful(Unauthorized(errorHandler.internalServerErrorTemplate))) { id =>
-      identityVerificationConnector.verifySuccess(id).flatMap {
-        case true =>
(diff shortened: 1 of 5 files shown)


app/controllers/IdentityVerification.scala
```diff
@@ -76,15 +76,19 @@ class IdentityVerification @Inject()(
...
+          case true =>
+            identityVerificationService.continue(id, request.userDetails).map {
+              case Some(obj) => identityVerificationService.someCase(obj)
+              case None      => identityVerificationService.noneCase
+            }
+          case false => Future.successful(Unauthorized(errorHandler.internalServerErrorTemplate))
+        }
       }
+    } else {
+      Future.successful(Redirect(controllers.registration.routes.RegistrationController.success(1L)))
```
darrell-plant: 1L is a success code?

darrell-plant:
LGTM 👍

app/controllers/IdentityVerification.scala
PeterPerhac: no, it's just some person id gets displayed on a confirmation screen. can't dig that out of nowhere since this is iv success screen but we have switched IV off...

DanielConnelly:
LGTM 👍

PR #793, opened 2020-04-03
[VTCCA-3344] - fixed file upload error message issue

jon-pike:


PeterPerhac:
👍 LGTPP

app/assets/javascripts/src/fileUpload.js
```diff
@@ -19,10 +19,13 @@
         $element.change(function(){
             $('#error-summary').remove();
             $('#uploadFile').attr('disabled','disabled');
+            var file = this.files[0];
 
-            $(this).after('<div class="message-warning" id="message-warning"><p>Please wait whilst your file is uploading. This may take some time.</p></div>');
 
-            var file = this.files[0];
+            if(file.type != "" && file.type != undefined){
```
darrell-plant, line 25: you could try this: https://stackoverflow.com/questions/5515310/is-there-a-standard-function-to-check-for-null-undefined-or-blank-variables-in

darrell-plant:
otherwise LGTM 👍

app/assets/javascripts/src/fileUpload.js
jon-pike: Thanks Darrell - works a treat

PR #802, opened 2020-04-15
Vtcca 3208: 

habeeb01:


reddy-yattapu:
:+1:

app/actions/requests/BasicAuthenticatedRequest.scala
```diff
@@ -36,11 +36,25 @@ case class BasicAuthenticatedRequestWithAffinityGroup[A](
...
       request: Request[A]
-) extends AuthenticatedRequest[A](request)
+) extends AuthenticatedRequest[A](request) with CcaWrappedRequest {
+  override def optAccounts: Option[Accounts] =
+    Some(Accounts(organisationAccount, individualAccount))
+
+  override def isLoggedIn: Boolean = optAccounts.nonEmpty
+
+  override def yourDetailsName: Option[String] = optAccounts.map { acc =>
+    if (s"${acc.person.details.firstName} ${acc.person.details.lastName}" == acc.organisation.companyName) {
```
PeterPerhac, line 51: this `${acc.person.details.firstName} ${acc.person.details.lastName}` appears here three times.
would be easier to see what's going on if you assigned it to a val
`val fullName = s"${acc.person.details.firstName} ${acc.person.details.lastName}"` and then used `fullName` in the three places

PeterPerhac:
minor nitpick otherwise looks good

app/actions/requests/BasicAuthenticatedRequest.scala
```diff
@@ -36,11 +36,25 @@ case class BasicAuthenticatedRequestWithAffinityGroup[A](
...
+      s"${acc.person.details.firstName} ${acc.person.details.lastName}"
+    } else {
+      s"${acc.person.details.firstName} ${acc.person.details.lastName} - ${acc.organisation.companyName}"
+    }
+  }
```
darrell-plant, line 56: This code should be in CaWrappedRequest as it is in vmv-frontend. Did you remove from `AppointAgentSessionRequest` ?

habeeb01: no I didn't remove AppointAgentSessionRequest. AppointAgentSessionRequest carries the session data required for the add agent journey. The manage agent pages don't use AppointAgentSessionRequest, they don't have any session data saved between pages. It only needs the authenticated user's details

darrell-plant: I think you misunderstood. The implementation of `yourDetailsName` is in the parent trait CCaWrappedRequest in vmv-frontend (where you took it from originally). Previously you had an identical implementation of `yourDetailsName` in AppointAgentSessionRequest.

I'm asking if you have removed that implementation from AppointAgentSessionRequest (otherwise it's duplicated)? And I'm suggesting moving this implementation to CCaWrappedRequest - as it is in vmv-frontend

darrell-plant:
Are you sure about passing the backLinks around via forms? In other places we use the Referer header to determine the back link and that seems to work. Did you consider that? Not sure if it is applicable in your case, but worth considering...

app/actions/requests/BasicAuthenticatedRequest.scala
habeeb01: yep, I definitely misunderstood you. I agree with your idea

habeeb01:
We have had issues with the referer where users end up in a loop and when users' session have died. Apart from passing the backLink to the page, I don't see any other way. Im open to ideas

darrell-plant:
> We have had issues with the referer where users end up in a loop and when users' session have died. Apart from passing the backLink to the page, I don't see any other way. Im open to ideas

Just mentioning referer in case you hadn't considered it.

darrell-plant:
LGTM 👍

PR #809, opened 2020-04-22
VTCCA-3209: option should now be highlighted after clicking on error …

habeeb01:
…message


jon-pike:
👍

darrell-plant:
Wonder why removing ids from radio elements changes behaviour like that?
Anyway, LGTM 👍

PR #820, opened 2020-05-05
VTCCA-3422: removing unused code and new agent relationship flag

habeeb01:


app/controllers/agent/ManageAgentController.scala
```diff
@@ -313,7 +313,7 @@ class ManageAgentController @Inject()(
         params = GetPropertyLinksParameters(
           sortfield = ExternalPropertyLinkManagementSortField.ADDRESS,
           sortorder = ExternalPropertyLinkManagementSortOrder.ASC),
-        agentCode = agentCode,
-        backLink = controllers.agent.routes.ManageAgentController.manageAgent(Some(agentCode)).url
-      ))
+        agentCode = agentCode
```
darrell-plant, line 316: why did the backLink go?

habeeb01: now we can only get to this particular page from one place so no need to pass it into the controller

app/views/dashboard/managedByAgentsProperties.scala.html
```diff
@@ -91,11 +91,7 @@ <h2 id="noAgentId" class="heading-secondary">
                         <td>
                             <ul class="list">
                                 <li>
-                                    <a href="@agentAppointment.routes.RevokeAgentController.revokeAgent(
-                                        p.submissionId,
-                                        p.agents.filter(ag =>
-                                            Some(ag.organisationId) == model.agentOrganisationId).head.authorisedPartyId,
-                                        model.agentCode)">
+                                    <a href="#">
```
darrell-plant, line 94: should this still be an anchor? maybe remove that?

darrell-plant:
couple of comments but LGTM 👍

app/views/dashboard/managedByAgentsProperties.scala.html
habeeb01: this page not used anymore, it is from when we had the old dashboard (when Fi was still in the team). I didn't remove it because I wanted this change to just be about the new agent relationship stuff and the issue we found the other day. We can remove this file (and the old Dashboard code) in another ticket. There is some redirect logic in the old Dashboard that we will need to handle/keep as well.

reddy-yattapu:
:+1:

reddy-yattapu:
:+1:

app/views/dashboard/managedByAgentsProperties.scala.html
darrell-plant: well, it would hurt to remove the tags <a ...>...</a>

darrell-plant: but you have my thumb

PeterPerhac:
👍

PR #821, opened 2020-05-07
VTCCA-3412: was changing state with GET, updated that to a POST and u…

habeeb01:
…sing new design systems for revoke client

### DIFF
--- a/conf/app.routes
+++ b/conf/app.routes
@@ -31,7 +31,7 @@ GET         /managed-properties/:agentCode                                      
 GET         /manage-agents                                                                    @controllers.Dashboard.manageAgents()
 
 GET         /client-properties/revoke/:plSubmissionId                                         @controllers.agent.RepresentationController.revokeClient(plSubmissionId: String)
-GET         /client-properties/revoke-client-property/:plSubmissionId/confirmed               @controllers.agent.RepresentationController.revokeClientPropertyConfirmed(plSubmissionId: String)
+POST        /client-properties/revoke-client-property/:plSubmissionId/confirmed               @controllers.agent.RepresentationController.revokeClientPropertyConfirmed(plSubmissionId: String)
 
 GET         /property-link/:submissionId/assessments                                          @uk.gov.hmrc.propertylinking.controllers.valuations.ValuationsController.valuations(submissionId: String, owner: Boolean ?= true)
 GET         /property-link/clients/all/:submissionId/assessments                              @uk.gov.hmrc.propertylinking.controllers.valuations.ValuationsController.valuations(submissionId: String, owner: Boolean ?= false)
--- a/test/controllers/agent/RepresentationControllerSpec.scala
+++ b/test/controllers/agent/RepresentationControllerSpec.scala
@@ -23,7 +23,7 @@ import org.mockito.Mockito.when
 import org.scalacheck.Arbitrary.arbitrary
 import play.api.mvc.AnyContentAsEmpty
 import play.api.test.FakeRequest
-import play.api.test.Helpers._
+import play.api.test.Helpers.{contentAsString, _}
 import play.twirl.api.Html
 import resources._
 import uk.gov.hmrc.http.HeaderCarrier
@@ -48,7 +48,7 @@ class RepresentationControllerSpec extends VoaPropertyLinkingSpec {
 
   behavior of "revokeClientPropertyConfirmed method"
   it should "revoke an agent and redirect to the client properties page" in {
-    when(revokeClientPropertyPage.apply(any())(any(), any(), any())).thenReturn(Html(""))
+    when(revokeClientPropertyPage.apply(any())(any(), any(), any())).thenReturn(Html("Revoking client"))
     val clientProperty: ClientPropertyLink = arbitrary[ClientPropertyLink]
 
     StubPropertyLinkConnector.stubClientPropertyLink(clientProperty)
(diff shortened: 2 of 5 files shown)


conf/app.routes
```diff
@@ -31,7 +31,7 @@ GET         /managed-properties/:agentCode
 GET         /manage-agents                                                                    @controllers.Dashboard.manageAgents()
 
 GET         /client-properties/revoke/:plSubmissionId                                         @controllers.agent.RepresentationController.revokeClient(plSubmissionId: String)
-GET         /client-properties/revoke-client-property/:plSubmissionId/confirmed               @controllers.agent.RepresentationController.revokeClientPropertyConfirmed(plSubmissionId: String)
+POST        /client-properties/revoke-client-property/:plSubmissionId/confirmed               @controllers.agent.RepresentationController.revokeClientPropertyConfirmed(plSubmissionId: String)
```
darrell-plant, line 34: should it be a PATCH?

test/controllers/agent/RepresentationControllerSpec.scala
```diff
@@ -57,15 +61,13 @@ class RepresentationControllerSpec extends VoaPropertyLinkingSpec {
 
   behavior of "revokeClient method"
   it should "revoke an agent and display the revoke client page" in {
+    when(revokeClientPropertyPage.apply(any())(any(), any(), any())).thenReturn(Html(""))
```
darrell-plant, line 64: Was there a reason not to render the actual page and test against it?

conf/app.routes
habeeb01: why not POST?

darrell-plant: Maybe we don't care about that on the frontend?

darrell-plant: No we don't. It's okay.

test/controllers/agent/RepresentationControllerSpec.scala
habeeb01: I don't think it we should be checking for html in unit tests (this is an old). The acceptance tests can do that.

darrell-plant: That wasn't Dan's advice.

darrell-plant: I guess there's not much logic on this page, so a single acceptance test would be enough. But if there is logic leading to different behaviour on the page, then unit tests are useful as we generally don't have acceptance tests for all possibilities, and unit tests catch errors sooner.

darrell-plant: We have used jsoup in recent pages to parse the page and then make various tests.

habeeb01: This is what we've done is other frontend stories.

What was Dan's advice?

darrell-plant: One of Dan's handovers talked about developing a consistent pattern for testing pages - and we were using Jsoup for this. I'm saying that we have to do this way, but we should have a consistent approach so we are all on the same page (no pun intended).

habeeb01: I think this is one that we might need to discuss as a dev team (so we can agree how we do this going forward). 

Seeing as there isn't any logic on this page, are you happy to give your thumb?

darrell-plant: why don't you remove "when(revokeClientPropertyPage.apply(any())(any(), any(), any())).thenReturn(Html(""))" 

this is not a good idea

habeeb01: I think it is perfectly valid if we are not testing the html. We are just making sue that the mocked page loads, we are not interested in what is on the mocked page. Yes lets discuss for sure

darrell-plant: yes, but ... why? :)

darrell-plant: "if we are not testing the html" - you removed the test and I'm questioning it :)

darrell-plant: and returning empty html instead of a real page may hide errors - it's just about whether we test the content

test/controllers/agent/RepresentationControllerSpec.scala
```diff
@@ -57,6 +61,8 @@ class RepresentationControllerSpec extends VoaPropertyLinkingSpec {
 
   behavior of "revokeClient method"
   it should "revoke an agent and display the revoke client page" in {
+    when(revokeClientPropertyPage.apply(any())(any(), any(), any()))
+      .thenReturn(Html("Revoking client <br/> Are you sure you no longer want to act on behalf of"))
```
darrell-plant, line 65: This is not right. This was not here in the original. I'm questioning why you are mocking the page?
Of course the content checks you've added are pointless if you've mocked the page, but you shouldn't need to mock the page if you've mocked your services correctly.

Sorry, this is worse than before.

habeeb01: Lets stay professional please

Ok i think I see where you are getting confused. In the play 2.6 version of design systems, you have to inject the html/twirl files, because we are doing so, we need to provide a mock in the tests. Otherwise we just get null pointers. If you have a nicer way around this then please enlighten me

darrell-plant: "Lets stay professional please" - no idea why you are saying that.

darrell-plant: Are you saying there is technical reason why we can't render real pages in out unit tests now? Or that we are choosing not to do it?

habeeb01: I don't know how you would render the real page, and I don't think we need to, especially if there is no logic to test on the page

darrell-plant: If you are saying that we can't do it, then the empty html is obviously better than the latest one - which is why I said it is 'worse than before'.

darrell-plant: > I don't know how you would render the real page, and I don't think we need to, especially if there is no logic to test on the page

Well, we'll end up going in circles. But what if we do want to test the page. Are you saying we can't in 2.6?

habeeb01: Im saying, as far as Im aware, you would have to mock the behaviour you want unless you render the real page. I don't know how to do that and there is no point in this case. We can discuss a general case (where we want to test for some logic) separately. It should not need to hold up this PR. Thats really all im trying to say.

darrell-plant: Agreed that for this page just mock the empty page - I didn't know it was technically difficult. Please revert your last commit and I'll thumb that.

But we need to discuss this as my understanding was that we should be testing pages where there is logic involved.

btw this is a discussion where we can all gain an understanding of what is going on.

darrell-plant:
LGTM 👍

jon-pike:
👍

PR #839, opened 2020-08-17
VTCCA-3166: fixed back link issue found in testing

habeeb01:


app/uk/gov/hmrc/propertylinking/controllers/valuations/ValuationsController.scala
```diff
@@ -74,24 +75,44 @@ class ValuationsController @Inject()(
...
-                ),
-                owner = owner
-              ))
+          if (owner) {
+            Future.successful(
+              Ok(
+                views.html.dashboard.assessments(
+                  model = AssessmentsVM(
+                    assessmentsWithLinks = link.assessments
+                      .sortBy(_.currentFromDate.getOrElse(LocalDate.of(2017, 4, 7)))(
```
darrell-plant, line 84: We should factor this date out. Maybe config?

app/uk/gov/hmrc/propertylinking/controllers/valuations/ValuationsController.scala
```diff
@@ -137,7 +158,9 @@ class ValuationsController @Inject()(
       clientDetails match {
         case Some(client) =>
           config.newDashboardUrl(
-            s"selected-client-properties?clientOrganisationId=${client.organisationId}&clientName=${client.organisationName}&pageNumber=1&pageSize=15&sortField=ADDRESS&sortOrder=ASC")
+            s"selected-client-properties?clientOrganisationId=${client.organisationId}&clientName=${URLEncoder
+              .encode(client.organisationName, "UTF-8")}&pageNumber=1&pageSize=15&sortField=ADDRESS&sortOrder=ASC")
```
darrell-plant, line 162: are these the defaults (&pageNumber=1&pageSize=15&sortField=ADDRESS&sortOrder=ASC)
if so, we don't need them here.
I'm not sure if they are, but might simplify that code

darrell-plant:
LGTM 👍 (with comments)

app/uk/gov/hmrc/propertylinking/controllers/valuations/ValuationsController.scala
habeeb01: yes they are defaults but unfortunately the link doesn't seem to work without them 🤷‍♂️ . That is the only reason why I added them

app/uk/gov/hmrc/propertylinking/controllers/valuations/ValuationsController.scala
habeeb01: Good point. I will do this in a separate commit

PR #840, opened 2020-08-20
VTCCA-3166: fixed back link issue in assessments page

habeeb01:


conf/app.routes
```diff
@@ -33,7 +33,9 @@ GET         /manage-agents
 GET         /client-properties/revoke/:plSubmissionId                                         @controllers.agent.RepresentationController.revokeClient(plSubmissionId: String)
 POST        /client-properties/revoke-client-property/:plSubmissionId/:address/confirmed      @controllers.agent.RepresentationController.revokeClientPropertyConfirmed(plSubmissionId: String, address: String)
 
+GET         /property-link/:submissionId/assessments/start                                    @uk.gov.hmrc.propertylinking.controllers.valuations.ValuationsController.savePreviousPage(previousPage: String, submissionId: String, owner: Boolean ?= true)
```
reddy-yattapu, line 36: Just looking at URI and method name, it seems method name not sync with URI name, if possible think about matching method name for URI, otherwise you can ignore it

habeeb01: i don't think they need to match. That particular uri is just used for saving the previous page. I didn't want to mention "save-previous-page" in the url

reddy-yattapu:
👍

Leeroyc720 (commented):
LGTM👍

PR #845, opened 2020-08-28
VTCCA-3651: Scala 2.12 and SBT 1.3

PeterPerhac:


app/services/RegistrationService.scala
```diff
@@ -71,7 +71,7 @@ class RegistrationService @Inject()(
         option: Option[DetailedIndividualAccount],
         addressId: Long,
         groupAccount: Option[GroupAccount],
-        affinityGroupOpt: Option[AffinityGroup] = None)(
+        affinityGroupOpt: Option[AffinityGroup])(
```
darrell-plant, line 74: just double-checking you are sure about removing the default (None)?

darrell-plant:
LGTM 👍

app/services/RegistrationService.scala
PeterPerhac: of course, scala 2.12 can spot when the default value is never used. all callers explicitly provide this argument. neat

habeeb01:
👍 LGTM

PR #878, opened 2020-10-26
VTCCA-3728: DVR issue fix

darrell-plant:
Handling service call errors (use custom HttpReads over custom HttpErrorFunctions)


PeterPerhac:
👍  LGTPP I think I can follow what's going on there. I was sorting out something very similar some time ago. It's a tough task

conf/logback.xml
```diff
@@ -13,7 +13,7 @@
 
   <logger name="com.ning.http.client" level="${logger.com.ning.http.client:-INFO}"/>
 
-  <root level="DEBUG">
+  <root level="INFO">
```
habeeb01, line 16: is this on purpose?

habeeb01:
LGTM 👍  - apart from the comment re log level

conf/logback.xml
darrell-plant: yeah - I think I accidentally committed the DEBUG in my previous PR but that wasn't spotted - just resetting here.

PR #884, opened 2020-10-29
VTCCA-3728: restore app.routes to fix IV issue

darrell-plant:


PeterPerhac:
👍  shouldn't be needed , as the migration document says, but for now IV is broken and this fixes it in QA 👍

PR #894, opened 2020-11-17
VTCCA-3359: go to DVR if rateableValue is null

habeeb01:


darrell-plant:
LGTDz 👍

PeterPerhac:
i think this fix would like a unit test

PR #895, opened 2020-11-19
VTCCA-3359: added help text for when rateableValue is null in current…

habeeb01:
…Valuation


PeterPerhac:
👍  looks good to me

darrell-plant:
LGTDz 👍

PR #899, opened 2020-11-24
VTCCA-3786 Property Linking Start page Timeout Dialog Bug

reddy-yattapu:
Timeout Dialog should be shown only if user is logged in


PeterPerhac:
👍  lgtm

PR #911, opened 2021-01-07
VTCCA-3853: added validation to ensure phone number is all digits

habeeb01:


darrell-plant:
LGTDz 👍

app/models/registration/AdminUser.scala
```diff
@@ -81,7 +82,7 @@ object AdminUser {
       keys.address                -> addressMapping,
       keys.dateOfBirth            -> dmyPastDate,
       keys.nino                   -> nino,
-      keys.phone                  -> nonEmptyText(maxLength = 15),
+      keys.phone                  -> nonEmptyText(maxLength = 15).verifying("error.invalidPhoneNumber", num => num.forall(_.isDigit)),
```
PeterPerhac, line 85: i wonder if this is the correct validation rule, because i think it would be valid to include some spaces or the plus sign, or possibly some people might include dashes or brackets... 

`+44 (0)7591 231 233`

or something like that... could still be considered valid.

PeterPerhac:
It looks good to me but i wonder if it isn't a bit too strict. generally a regex is used to validate phone numbers, perhaps something like this https://stackoverflow.com/questions/11518035/regular-expression-for-gb-based-and-only-numeric-phone-number

app/models/registration/AdminUser.scala
habeeb01: fair point, I hadn't considered +'s or brackets. "+44(0)7951595104" is 16 characters (with no spaces). The max allowed for that field is 15. We'll need to agree what the validation should be for that field (maybe only + and numbers). I will bring it up at the stand up

PeterPerhac:
👍  LGTM

PR #924, opened 2021-02-04
VTCCA-3480 Claim Property journey change to adopt Welsh properties ba…

reddy-yattapu:
I had a chat with Karen
I find the First propertyLinkEarliestStartDate from Summary List, It does not matter It would be ListType is either Draft or Current

### DIFF
--- a/app/services/propertylinking/PropertyLinkingService.scala
+++ b/app/services/propertylinking/PropertyLinkingService.scala
@@ -72,7 +72,7 @@ class PropertyLinkingService @Inject()(
     for {
       propertyHistory <- propertyLinkConnector.getPropertyHistory(uarn)
       earliestStartDate: Option[LocalDate] = propertyHistory.history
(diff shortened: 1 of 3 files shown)


app/services/propertylinking/PropertyLinkingService.scala
```diff
@@ -72,7 +72,7 @@ class PropertyLinkingService @Inject()(
     for {
       propertyHistory <- propertyLinkConnector.getPropertyHistory(uarn)
       earliestStartDate: Option[LocalDate] = propertyHistory.history
-        .find(pv => pv.listType == ListType.CURRENT && pv.valuationStatus == ValuationStatus.CURRENT)
+        .find(_.propertyLinkEarliestStartDate.isDefined)
```
habeeb01, line 75: is there a test that covers this scenario? might be worth adding one if it doesn't already exist

reddy-yattapu: I have added new testcase Please check

habeeb01:
LGTM 👍

PR #926, opened 2021-02-23
VTCCA-3884 - Page title has '- Valuation Office Agency - GOV.UK' twice

zahiriqbal:
Removed duplicate wording from messages file


zahiriqbal:
Error in signing

PR #933, opened 2021-03-02
VTCCA-3912 changed the options for a question

zahiriqbal:


darrell-plant:
LGTDz 👍 (have agreed for now not to change the error message as suggested by Rhys as we use a generic message of "Please select an option" on simple radio groups. If we need custom messages per page, then it could be done as part of a bigger change or folding into this PR after discussion with @habeeb-olayemi )

habeeb01:
I agree with you @darrell-plant. We'll probably need to feed that back to Rhys. Just so he's aware of the content that will be released (we can mention at the stand up)....LGTM 👍

PR #934, opened 2021-03-04
VTCCA-3954 Migrate to External Available Property Links endpoint

reddy-yattapu:


habeeb01:
LGTM 👍

app/connectors/propertyLinking/PropertyLinkConnector.scala
```diff
@@ -103,29 +103,27 @@ class PropertyLinkConnector @Inject()(config: ServicesConfig, http: HttpClient)(
...
-        Some("sortorder"                    -> searchParams.sortorder.toString),
-        agentAppointed.map("agentAppointed" -> _),
-        Some("organisationId"               -> organisationId.toString),
-        Some("agentOrganisationId"          -> agentOrganisationId.toString)
-      ).flatten ++
-        List(
-          "startPoint"           -> pagination.startPoint.toString,
-          "pageSize"             -> pagination.pageSize.toString,
-          "requestTotalRowCount" -> pagination.requestTotalRowCount.toString
+    val ownerAuthResult = agentAppointed.getOrElse("BOTH") match {
```
PeterPerhac, line 109: "BOTH" could be replaced by "foobar" and it would still work. I think in this case an if-statement would be better. 

```scala
if (agentAppointed.contains("NO") { getMyOrganisationsPropertyLinks(searchParams, pagination) } else {...}
```

app/connectors/propertyLinking/PropertyLinkConnector.scala
```diff
@@ -103,29 +103,27 @@ class PropertyLinkConnector @Inject()(config: ServicesConfig, http: HttpClient)(
...
+      case "NO" =>
+        getMyOrganisationsPropertyLinks(searchParams, pagination)
+      case _ =>
+        http.GET[OwnerAuthResult](
+          s"$baseUrl/my-organisation/agents/$agentCode/available-property-links",
+          List(
```
PeterPerhac, line 115: You're wrapping half of the mandatory parameters in Some. Then flattening. Either do all of them in Some and flatten at the end, or - since it's only the address that is optional - so you could this instead:
```scala
searchParams.address.map("address" -> _).toList ++ List(
 "sortField"            -> searchParams.sortfield.toString,
 "sortOrder"            -> searchParams.sortorder.toString,
 "startPoint"           -> pagination.startPoint.toString,
 "pageSize"             -> pagination.pageSize.toString,
 "requestTotalRowCount" -> pagination.requestTotalRowCount.toString
)
```

PeterPerhac:
2 small nitpicks but otherwise looks good 👍

PeterPerhac:
👍  Thanks

PR #942, opened 2021-04-06
VTCCA-3368: draft list DVR pages

PeterPerhac:


habeeb01:
LGTM 👍 - I would like to understand why we needed to change "checksAndChallenges" into a tuple. I don't have problem with the code but I would like to know why.

darrell-plant:
LGTDz 👍 (I struggled with trying to understand when we dealing with property-links and valuations as the naming is a bit confusing in places. Also, still not sure what a `ApiAssessments` is. But that's not a comment on this PR - more that area of code in general.)

PeterPerhac:
> LGTM 👍 - I would like to understand why we needed to change "checksAndChallenges" into a tuple. I don't have problem with the code but I would like to know why.

The pair of case details has become optional. They will either BOTH be there, or not. We're not fetching the case details on the DVR screen when user is viewing a draft list assessment

PR #956, opened 2021-05-17
VTCCA-4055 - Migrated view and updated controller and tests

WitchellJack:


app/views/propertyrepresentation/revokeAgentSummary.scala.html
```diff
@@ -18,28 +18,23 @@
 @import binders.pagination._
 @import binders.propertylinks._
 @import config.ApplicationConfig
+@this(mainLayout: views.html.mainLayout)
```
darrell-plant, line 21: Can we get a new line between lines 20 & 21, and 21 & 22, please?

darrell-plant:
LGTDz 👍 (minor comment around formatting)

app/views/propertyrepresentation/revokeAgentSummary.scala.html
WitchellJack: Done 👍🏻

app/views/propertyrepresentation/revokeAgentSummary.scala.html
```diff
@@ -18,28 +18,25 @@
...
 )(implicit request: AuthenticatedRequest[_], messages: Messages, config: ApplicationConfig)
 
 @heading = @{ Messages("propertyRepresentation.agentRemoveSummary.heading") }
 
-@main_template(
-heading,
-contentHeaderText = Some(heading),
-contentHeaderHidden = true
+@mainLayout(
+ heading,
```
PeterPerhac, line 32: here it would be good to explicitly specify the name of the parameter

`pageTitle = heading`

app/views/propertyrepresentation/revokeAgentSummary.scala.html
```diff
@@ -18,28 +18,25 @@
...
-
- <p class="lede" id="mainBodyText">@messages("propertyRepresentation.agentRemoveSummary.removed", action.agentCode.toString, action.propertyLinkIds.size)</p>
-
+ <h1 class="govuk-heading-l">@messages("propertyRepresentation.agentRemoveSummary.heading")</h1>
 
- <a class="button" id="selectMorePropertiesButton" href="@agentAppointment.routes.AppointAgentController.selectAgentPropertiesSearchSort(PaginationParameters(), action.agentCode)" >@messages("propertyRepresentation.agentSummary.select.more")</a>
+ <p class="govuk-body" id="mainBodyText">@messages("propertyRepresentation.agentRemoveSummary.removed", action.agentCode.toString, action.propertyLinkIds.size)</p>
 
-</div>
+ <a class="govuk-button" id="selectMorePropertiesButton" href="@agentAppointment.routes.AppointAgentController.selectAgentPropertiesSearchSort(PaginationParameters(), action.agentCode)" >@messages("propertyRepresentation.agentSummary.select.more")</a>
```
PeterPerhac, line 40: didn't the link look a bit smaller than usual in browser? I found that I had to put links into `govuk-body` p-tags for the links to have the right size.

PeterPerhac:
looks good to me 👍  (just a minor nitpick and a question)

app/views/propertyrepresentation/revokeAgentSummary.scala.html
PeterPerhac: oh, i see it's "govuk-button" type of link... in that case shouldn't it be a govukButton component? 

something like this... ( from my own branch) 

`@govukButton(Button(content = Text(messages("invalidAccountType.register")), href = Some(routes.Register.show().url)))`

WitchellJack: So I am happy to go with either, I had this originally as a govUK button but amended it after a side discussion with Habeeb, it seemed to make sense to simply update the "button" to a "govUkButton" instead of introducing a new compoment, but also happy to revert it back to the GovUkButton - @habeeb-olayemi  what do you think?

habeeb01: its a link that is styled as a button so I think what you have done is correct. I don't see a need to pull in the. govUkButton component

habeeb01:
LGTM 👍

app/views/propertyrepresentation/revokeAgentSummary.scala.html
PeterPerhac: I'll ask in plat UI channel, cuz I'd like to know. From design systems docs I can't see any use of a button-styled link.

WitchellJack: Okay cool, I'll hold fire on this for now

PeterPerhac: image

PeterPerhac: i think the govukButton component produces an `<a>` tag in the end, but the difference is that they can change the library to style the link differently and our button-styled links will automatically be updated if we pull in new version of design system library

WitchellJack: Updated back to button 👍🏻

PeterPerhac:
👍  LGTPP

PR #957, opened 2021-05-17
VTCCA-4050: migrate error pages

PeterPerhac:


app/views/errors/technicalDifficulties.scala.html
```diff
@@ -18,22 +18,22 @@
...
+    <h1 class="govuk-heading-xl">@messages("global.error.InternalServerError500.heading")</h1>
 
-    <p>@messages("internalServerError.p.2")</p>
+    <p class="govuk-body">@messages("internalServerError.p.1")</p>
+    <p class="govuk-body">@messages("internalServerError.p.2")</p>
 
-    @reference.map { r =>
-        <p>@messages("internalServerError.reference", r)</p>
+    @for(ref <- reference){
+        <p class="govuk-body">@messages("internalServerError.reference", ref)</p>
```
darrell-plant, line 33: Out of interest, why prefer `for` over `map`?

app/views/errors/error.scala.html
```diff
@@ -16,12 +16,15 @@
 
 @import config.ApplicationConfig
 
-@(title: String, heading: String, message: String)(implicit request: RequestHeader, messages: Messages, config: ApplicationConfig)
+@this(mainLayout: views.html.mainLayout)
 
 
-    @main_template(title = title, topNavigation = false) {
+@(title: String, heading: String, message: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
```
darrell-plant, line 22: In the templates, we have a mixture of 
`(implicit request: RequestHeader,...`
and
`(implicit request: Request[_],...`

is it better we standardise and stick to one (RequestHeader)?

darrell-plant:
LGTDz 👍 (with a couple of questions)

app/views/errors/technicalDifficulties.scala.html
PeterPerhac: it's a nice pattern for dealing with Option type in Twirl templates. Looks cleaner. A `map` actually maps the Option[A] to an Option[B] and the result is not needed. Instead of `map` a `foreach` should be used as it's only doing a side-effect of writing some HTML to the output. I believe the `for` comprehension will eventually end up a `foreach` because there's no `yield` keyword. It's just nicer like this, and I read somewhere that this is the recommended way to optionally print things in a twirl template.

app/views/errors/error.scala.html
PeterPerhac: every Request extends a RequestHeader. In ErrorHandler we have to override methods that have RequestHeader in the signature and I can't magic up a Request from that. Had to update to RequestHeader in the places where it didn't compile. There may still be a few where a Request is asked for, but that can be updated to RequestHeader...

reddy-yattapu:
:+1:

PR #960, opened 2021-06-08
VTCCA-4053-migrate-register-pages

zahiriqbal:


app/assets/javascripts/src/addressLookup.js
```diff
@@ -22,14 +22,26 @@
...
+            if($('.govuk-error-summary').length) {
                 showFields();
             }
         }
 
         function showLookupError() {
-            $('#postcodeSearchGroup').find('.error-message').remove();
-            $('#postcodeSearch').before('<p class="error-message">' + messages.errors.postcodeLookupError + '</p>').closest('.form-group').addClass('error');
+
+            var isContains = $('#postcodeSearchOnly').text().indexOf('Enter a valid postcode') > -1;
```
darrell-plant, line 32: `isContains` contains what? do you mean something like `isError`?

app/assets/javascripts/src/addressLookup.js
```diff
@@ -22,14 +22,26 @@
...
+
+    console.log(isContains)
```
darrell-plant, line 34: remove console logging

app/assets/javascripts/src/addressLookup.js
```diff
@@ -22,14 +22,26 @@
...
+    if(isContains == true){
```
darrell-plant, line 35: this is just ```if (contains)``` or ```if (isError)``` if you rename

app/views/createAccount/register_assistant_admin.scala.html
```diff
@@ -20,83 +20,122 @@
...
+@getEnrichedErrorLinks = @{
+model.errors.map{ error =>
+ErrorLink(href = Some(s"#${error.key}"),
+content = HtmlContent(s"${messages(s"label.${error.key}")} - ${messages(error.message)}") )}
+}
+@getFromDateFiledErrorMessage = @{
+(model.error("dob.day"), model.error("dob.month"), model.error("dob.year")) match
+{
+case (None, None, None ) => None
+case _                   => Some(ErrorMessage(content = Text(messages("error.common.invalid.date"))))
```
darrell-plant, line 43: shame about the formatting - I know not easy to do without the plugin

app/views/helpers/addressNew.scala.html
```diff
@@ -0,0 +1,131 @@
+@*
```
darrell-plant, line 1: we have to remember to rename this back to `address.scala.html` after Jack's ticket!

conf/application.conf
```diff
@@ -264,7 +264,7 @@ google-analytics {
 sign-out = {
   url = "/bas-gateway/sign-in"
   keep-alive-url = "/business-rates-property-linking/keep-alive"
-  timeout = 890
+  timeout = 89000
```
darrell-plant, line 267: reset this to 890!!

conf/messages
```diff
@@ -960,6 +971,7 @@ label.status=Status
 label.search=Search
 label.clear.search=Clear Search
 
+label.Address=<h1 class="govuk-heading-l">Address<h1/>
```
darrell-plant, line 974: why do we have html tags in here? They should be in the twirl template

conf/messages
```diff
@@ -1289,29 +1301,37 @@ create.success.personId=VOA Personal ID:
...
-enrolment.lastName = Last name
+label.mobilePhone = Mobile number
+optional.label = Optional
+enrolment.tradeName = <h1 class="govuk-heading-m">Trading name (Optional)</h1>
+enrolment.mobphone = <h1 class="govuk-heading-m">Mobile number</h1>
+enrolment.phone = <h1 class="govuk-heading-m">Telephone number</h1>
+enrolment.confirmedEmail = <h1 class="govuk-heading-m">Confirm email address</h1>
+enrolment.email = <h1 class="govuk-heading-m">Email address</h1>
+enrolment.firstName = <h1 class="govuk-heading-m">First name</h1>
+enrolment.lastName = <h1 class="govuk-heading-m">Last name</h1>
```
darrell-plant, line 1318: why do we have html tags in here? They should be in the twirl template
(and the others...)

conf/messages
zahiriqbal: @darrell-plant its used like this:  <th class="govuk-table__header">@messages("enrolment.organisation.name.label")</th> 

so I thought i'll put it in the message. 

Please let me know if this is not correct

conf/messages
zahiriqbal: Similar to the last ones i'm passing them like this: label = Label(content = HtmlContent(messages("enrolment.firstName"))),

@darrell-plant

conf/messages
darrell-plant: not correct :)

darrell-plant: also, follow existing conventions
`label.Address` should be `label.address`

conf/messages
WitchellJack: I've noticed some of our services use this method of having HTML within the Messages file - my last ticket I went through and removed the HTML which I was using. I think the Messages file should just be for plain text, although this does work and is done elsewhere, in my opinion we should keep our messages file strictly text only.

darrell-plant:
LGTDz 👍

app/views/createAccount/registration_confirmation.scala.html
```diff
@@ -22,26 +22,32 @@
...
-            <p>@messages("create.success.keepUsingGg")</p>
+@(id: String, affinityGroup: AffinityGroup, credentialRole: CredentialRole)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
 
-            @if(affinityGroup != Individual && credentialRole != Assistant) {
-                <p>@Html(Messages("account.confirmation.addToGG", routes.Application.addUserToGG()))</p>
-            }
+@mainLayout(pageTitle = messages("account.confirmation.title"), mainClass = Some("govuk-grid-column-full")) {
 
-            <p>@Html(Messages("account.confirmation.disclaimer"))</p>
+<div class="grid-row">
```
habeeb01, line 34: you don't need these 2 outer divs

app/views/createAccount/register_assistant.scala.html
```diff
@@ -21,72 +21,94 @@
...
+
+@getEnrichedErrorLinks = @{
+    model.errors.map{ error =>
+    ErrorLink(href = Some(s"#${error.key}"),
+    content = HtmlContent(s"${messages(s"label.${error.key}")} - ${messages(error.message)}") )}
+}
+
+@mainLayout(pageTitle = messages("assessments.title"), mainClass = Some("govuk-grid-column-full")) {
+
+<div class="grid-row">
```
habeeb01, line 43: you don't need these 2 outer divs (same in the other pages)

app/views/createAccount/register_organisation.scala.html
```diff
@@ -19,138 +19,197 @@
...
-                @inputText(
-                    field = model(keys.email),
-                    '_label -> Messages("enrolment.emailAddress"),
-                    '_help -> Messages("enrolment.emailAddress.hint"),
-                    '_labelClass -> "form-label-bold",
-                    '_showConstraints -> false,
-                    'class -> "form-control",
-                    '_value -> organisationFieldData.email
-                )
+    <div class="govuk-body">
```
habeeb01, line 48: you don't need these 2 outer divs

app/views/createAccount/register_assistant.scala.html
darrell-plant: good point!

habeeb01: @zahiriqbal after you have removed the 2 outer divs...do not pass "mainClass" tp mainLayout, mainLayout is already "govuk-grid-column-two-thirds" by default

darrell-plant: unless he needs `govuk-grid-column-full` 🤷 (as he was using both previously I'm not sure which one was actually applied)

habeeb01: i think the page was previously "column-two-thirds" so I'll assume we are not looking to change that

darrell-plant: yeah - should keep the original 👍 (just a bit confused by Z's layout :) )

habeeb01:
LGTM 👍

PR #961, opened 2021-06-08
Vtcca 4057 - Migrate update details pages

WitchellJack:


app/assets/javascripts/src/addressLookup.js
```diff
@@ -28,8 +28,19 @@
         }
 
         function showLookupError() {
-            $('#postcodeSearchGroup').find('.error-message').remove();
-            $('#postcodeSearch').before('<p class="error-message">' + messages.errors.postcodeLookupError + '</p>').closest('.form-group').addClass('error');
+            var isContains = $('#postcodeSearchOnly').text().indexOf('Enter a valid postcode') > -1;
```
darrell-plant, line 31: this is a copy of Zahir's which I have commented on in his PR. I think in this case it would have been better to wait for Zahir's PR before raising this

darrell-plant: just make sure you keep yours inline with Zahir's changes

WitchellJack: have updated this now to match Zahirs 👍🏻

habeeb01:
LGTM 👍

app/views/details/updateAddress.scala.html
```diff
@@ -19,27 +19,35 @@
...
 @(model: UpdateDetailsVM)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
 
-@main_template(
-    Messages("updateAddress.title"),
-    topBackLink = true,
-    bottomBackLink = true,
-    backLinkUrl = Some(controllers.manageDetails.routes.ViewDetails.show().url)
+@mainLayout(
+    pageTitle = Messages("updateAddress.title"),
+    mainClass = Some("govuk-grid-column-two-thirds"),
```
darrell-plant, line 28: `govuk-grid-column-two-thirds` is the default css class in mainLayout, so this is not needed

PR #969, opened 2021-06-24
VTCCA-4164-rename

zahiriqbal:


darrell-plant:
Probably can delete `personalDetails_enrolment.scala.html`

darrell-plant:
LGTDz 👍

PR #987, opened 2021-09-07
VTCCA-4291 only call IV if user's confidence level is insufficient

darrell-plant:


app/controllers/registration/RegistrationController.scala
```diff
@@ -133,6 +137,21 @@ class RegistrationController @Inject()(
       )
   }
 
+  // TODO check IV enabled flag here
```
habeeb01, line 140: you haven't done your TODO

darrell-plant: yeah - was going to do as part of the uplift ticket

reddy-yattapu:
:+1:

app/controllers/registration/RegistrationController.scala
habeeb01: i guess it doesn't really matter because the flag is checked here: `IdentityVerification.startIv()`

habeeb01:
LGTM 👍

app/controllers/registration/RegistrationController.scala
darrell-plant: It would be better placed in RegistrationController, but it would impact tests so leaving that until next time

PR #999, opened 2021-10-11
VTCCA-4324 Fix Error messages follow GDS pattern

reddy-yattapu:


app/views/details/updateAddress.scala.html
```diff
@@ -18,17 +18,23 @@
...
 
 @mainLayout(
     pageTitle = Messages("updateAddress.title"),
     backUri = Some(controllers.manageDetails.routes.ViewDetails.show.url)
 ) {
-    <h1 class="govuk-heading-l">@messages("updateAddress.title")</h1>
 
+    @if(model.form.errors.nonEmpty) {
+        @govukErrorSummary(ErrorSummary(errorList = model.form.errors.asTextErrorLinks, title = Text(messages("errors.heading.title"))))
+    }
```
darrell-plant, line 35: I think this code block is repeated 8 times.

darrell-plant: `/views/helpers/errorSummary.sala.html` is probably not used anymore (you can check) - that's probably where this code should be.

reddy-yattapu: I think errorSummary.sala.html is old one with old styles, I am creating new errorSummay file with simple

PR #1006, opened 2021-10-13
VTCCA-4334 - Updated claim confirmation page

WitchellJack:


app/views/linkingRequestSubmitted.scala.html
```diff
@@ -20,7 +20,7 @@
 @import uk.gov.hmrc.govukfrontend.views.html.components._
 @this(govukPanel: GovukPanel, govukButton : GovukButton, mainLayout: views.html.mainLayout)
 
-@(model: RequestSubmittedVM)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
+@(model: RequestSubmittedVM, localAuthorityReference: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
```
WitchellJack, line 23: Just realised, it may be neater for me to update the RequestSubmittedVM to contain the localAuthorityReference and pass it through that way... what do you guys think worth the refactor?

darrell-plant: sounds sensible :thu

habeeb01:
LGTM 👍

PR #1009, opened 2021-10-15
VTCCA-4373: updated challenge form link for DVRs

DanSmith98:


DanSmith98:
good spot, I had only looked at criteria one and not scrolled down :/ I'll get it sorted

PR #1018, opened 2021-11-02
VTCCA-4430:make address camel case

zahiriqbal:


PeterPerhac:
👍  LGTM

PeterPerhac:
actually, I didn't spot it at first, but this is also missing the unit test for the formatter. you could add it in

darrell-plant:
LGTDz 👍 (after unit test added)

PeterPerhac:
👍  looks good

PR #1028, opened 2021-11-16
VTCCA-4447 add back previously deleted files

PeterPerhac:
fix broken "Select All" links and missing grey background


zahiriqbal:
👍

habeeb01:
LGTM 👍  - not sure if the `dataTable` class is required but maybe we can look at that separately

PR #1029, opened 2021-11-16
Vtcca 4451 challenge error screens

zahiriqbal:


darrell-plant:
LGTDz 👍

app/views/dvr/cannotRaiseChallenge.scala.html
```diff
@@ -16,23 +16,38 @@
...
+@this(mainLayout: views.html.mainLayout, govukButton: GovukButton)
 
-@(model: CanChallengeResponse, homePageUrl: String, authorisationId: Long, backLinkUrl: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
+@(model: CanChallengeResponse, address: String, localAuth: String, homePageUrl: String, authorisationId: Long, backLinkUrl: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
 
 @mainLayout(
     pageTitle = messages("cannotRaiseChallenge.title"),
     backUri = Some(backLinkUrl)
 ) {
+    <ul class="govuk-list">
```
PeterPerhac, line 29: 2 things here:
- don't we have this assessment header already in some Twirl template?
- English text should be pulled out of `messages` file, not hard-baked into Twirl template

app/controllers/detailedvaluationrequest/DvrController.scala
```diff
@@ -311,43 +314,51 @@ class DvrController @Inject()(
...
-              s"${config.serviceUrl}${controllers.detailedvaluationrequest.routes.DvrController.myClientsRequestDetailValuationCheck(plSubmissionId, assessmentRef, uarn).url}"
-          Redirect(
-            config.businessRatesValuationFrontendUrl(s"property-link/valuations/startChallenge?backLinkUrl=$returnUrl"))
-        }
-      case Some(response) =>
-        if (response.result) {
-          val party = if (isOwner) "client" else "agent"
-          Future.successful(
-            Redirect(config.businessRatesChallengeUrl(
+    val eventualPropertyHistory: Future[PropertyHistory] = vmvConnector.getPropertyHistory(uarn)
```
PeterPerhac, line 317: I can see there have been changes to what the controller does, however I don't find an accompanying change in the unit/integration tests. Can you add a unit test for this controller method?

PeterPerhac, line 317: I can see there have been changes to what the controller does, however I don't find an accompanying change in the unit/integration tests. Can you add a unit test for this controller method?

app/views/dvr/cannotRaiseChallenge.scala.html
PeterPerhac: we have this assessment header in check frontend, and it looks like this

```html

<div id="assessment-heading" class="govuk-grid-column-two-thirds">
    <ul class="govuk-list">
        <li id="billingAuthorityReference">
            <span class="govuk-body-s govuk-!-font-weight-bold">@Messages("assessment_heading.billingAuthorityReference"):  </span>
            <span class="govuk-body-s">@assessment.billingAuthorityCode/@assessment.billingAuthorityReferenceNumber</span>
        </li>
        <li id="address">
            <span class="govuk-body-s govuk-!-font-weight-bold">@Messages("assessment_heading.property"): </span>
            <span class="govuk-body-s">@assessment.fullAddress</span>
        </li>
    </ul>
</div>
```

it uses `span` with govuk css classes instead of `<b>` elements

zahiriqbal: @PeterPerhac  in PL theres only this used else where   <ul class="govuk-list">
        <li>
            <b>Local authority reference:</b> @model.localAuthorityReference
        </li>
        <li>
            <b>Property:</b> @Formatters.capitalisedAddress(model.address)
        </li>
    </ul>

PeterPerhac: we should not be using the `<b>` tag, instead we should be using a span with the appropriate govuk CSS classes (like we do in check).

about the b tag:
> This was formerly known as the Boldface element, and most browsers still draw the text in boldface. However, you should not use <b> for styling text; instead, you should use the CSS font-weight property to create boldface text, or the <strong> element to indicate that text is of special importance.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b

test/controllers/detailedvaluationrequest/DvrControllerSpec.scala
```diff
@@ -69,6 +67,132 @@ class DvrControllerSpec extends VoaPropertyLinkingSpec {
...
+    val result = resultCanChallenge(true)
+
+    val expectedRedirect =
+      s"http://localhost:9531/business-rates-challenge/property-link/$plSubmissionId/valuation/$assessmentRef/check/$caseRef/party/client/start?isDvr=true"
+
+    status(result) mustBe SEE_OTHER
+    redirectLocation(result) mustBe Some(expectedRedirect)
+  }
+
+  "can Challenge for Agent" must "redirect to start challenge when canChallenge response is true" in new CanChallengeSetup {
```
PeterPerhac, line 144: I would prefer `redirect to start challenge when the user can challenge` instead of `redirect to start challenge when canChallenge response is true` but it's really a minor thing. No one really ever reads this so I think it's ok to leave alone. Just for future tests, try to express what the test does in non-implementation-specific terms

PR #1030, opened 2021-11-17
VTCCA-4396 updated remove agent page layout and tests

DanSmith98:


app/views/propertyrepresentation/revokeAgentProperties.scala.html
```diff
@@ -129,165 +128,151 @@
...
+            ),
+            classes = "govuk-!-width-one-third"
+        ))
+
+        <div class="govuk-button-group">
+            @govukButton(Button(
+                content = Text(messages("label.search"))
+            ))
+            <a href="@agentAppointment.routes.AppointAgentController.selectAgentPropertiesSearchSort(PaginationParameters(), agentCode).url" class="govuk-body govuk-link">@messages("manageProperties.table.clear.search")</a>
+        </div>
```
darrell-plant, line 172: I think it is better to have the class `govuk-body` on the surrounding div, not on the <a>, if it is needed at all. I'm not sure, but that's would I would have done I think

test/controllers/agentAppointment/AppointAgentControllerSpec.scala
```diff
@@ -439,7 +439,7 @@ class AppointAgentControllerSpec extends VoaPropertyLinkingSpec with MockitoSuga
 
     val page = HtmlPage(Jsoup.parse(contentAsString(res)))
     page.mustContainText("Select one or more properties")
-    page.titleMustMatch("Error: Which properties should the agent be removed from - Valuation Office Agency - GOV.UK")
+    page.titleMustMatch("Error: Remove agent gg-ext-id from one or more properties - Valuation Office Agency - GOV.UK")
```
darrell-plant, line 442: I guess `gg-ext-id` is defined somewhere (in FakeObjects, as `ggExternalId`). In which case, this would be more stable:
```
page.titleMustMatch(s"Error: Remove agent $ggExternalId from one or more properties - Valuation Office Agency - GOV.UK")
```

test/controllers/agentAppointment/AppointAgentControllerSpec.scala
```diff
@@ -460,7 +460,7 @@ class AppointAgentControllerSpec extends VoaPropertyLinkingSpec with MockitoSuga
     status(res) mustBe BAD_REQUEST
     val page = HtmlPage(Jsoup.parse(contentAsString(res)))
     page.mustContainText("Failed to appoint agent to all properties")
-    page.titleMustMatch("Error: Which properties should the agent be removed from - Valuation Office Agency - GOV.UK")
+    page.titleMustMatch("Error: Remove agent gg-ext-id from one or more properties - Valuation Office Agency - GOV.UK")
```
darrell-plant, line 463: again

app/views/propertyrepresentation/revokeAgentProperties.scala.html
DanSmith98: I'll take a look at changing this and the tests, thanks

app/views/propertyrepresentation/revokeAgentProperties.scala.html
```diff
@@ -37,11 +37,11 @@
 )(implicit request: AuthenticatedRequest[_], messages: Messages, config: ApplicationConfig)
 
 @title = @{
-    Messages("propertyRepresentation.removeAgentProperties.title")
+    messages("propertyRepresentation.removeAgentProperties.heading", model.agentGroup.companyName)
```
PeterPerhac, line 40: so is it a title or a heading?
Heading is the `h1` tag, big text at top of the page. `title` is the text displayed in the browser tab. here we're assigning a heading to `@title` but sometimes `@title` is mis-named and is actually a heading... if it's a heading, can we change `@title` to `@heading` ?

conf/messages
```diff
@@ -419,8 +419,8 @@ propertyRepresentation.removeAgent.title=Remove agent permissions
 propertyRepresentation.removeAgent.message.p1=You can remove an agent''s ability to carry out checks and challenges on your behalf.
 propertyRepresentation.removeAgent.message.p2=Select an agent from the list below to remove them from one or more properties.
 propertyRepresentation.removeAgentProperties.title=Which properties should the agent be removed from
-propertyRepresentation.removeAgentProperties.heading=Remove an agent
-propertyRepresentation.removeAgentProperties.message.p1=You can remove {0}''s (Agent ID: {1}) ability to carry out checks and challenges for any of these properties:
+propertyRepresentation.removeAgentProperties.heading=Remove agent {0} from one or more properties
+propertyRepresentation.removeAgentProperties.message=(Agent ID: {0}) will no longer be able to submit checks and challenges for these properties.
```
PeterPerhac, line 423: `(Agent ID: {0})`  is this really the text that's supposed to show ? I find it strange that it's in round brackets and that we show some ID to the user, rather than the agent name... I'm just checking, I find the brackets especially strange

app/views/propertyrepresentation/revokeAgentProperties.scala.html
DanSmith98: They are the same thing in this case, so I used the similar existing .heading message for the title. You're right that it's ambiguous, if you feel it's better off as two duplicate messages for clarity, I can change it.

darrell-plant:
LGTDz 👍

conf/messages
DanSmith98: In the HTML, it has the agent name in bold, and uses this message afterwards, so the page does show the agent name and then the ID

app/views/propertyrepresentation/revokeAgentProperties.scala.html
darrell-plant: 'title' and 'heading' are supposed to be the same text on all our pages (except that 'title' has the 'VOA ...' suffix). I don't think we want/need two duplicate messages. I prefer 'title' (in the message file) as that is the non-technical term for the page - i.e. page title - the title then gets used as the browser tab title and the html page heading. But we don't seem to have an agreed convention for this.

darrell-plant: (I'm happy with the way it is in this PR)

PeterPerhac:
👍  looks good to me

PR #1035, opened 2021-11-19
VTCCA-4451: improve page title and spec as per req

zahiriqbal:


app/views/dvr/cannotRaiseChallenge.scala.html
```diff
@@ -23,7 +23,7 @@
 @(model: CanChallengeResponse, address: String, localAuth: String, homePageUrl: String, authorisationId: Long, backLinkUrl: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
 
 @mainLayout(
-    pageTitle = messages("cannotRaiseChallenge.title"),
+    pageTitle = messages("cannotRaiseChallenge.title.fail"),
```
PeterPerhac, line 26: I don't think you should add the `.fail` suffix. it is the cannotRaiseChallenge page, and this is its `title`. Why `.fail`?

zahiriqbal: @PeterPerhac Updated

darrell-plant:
LGTDz 👍

PeterPerhac:
👍  looks good to me

PR #1036, opened 2021-11-22
VTCCA-4396 add ids and names

DanSmith98:


app/views/propertyrepresentation/revokeAgentProperties.scala.html
```diff
@@ -55,7 +55,7 @@
         <input class="govuk-checkboxes__input" id="@id" name="linkIds[]" type="checkbox" value="@value">
         <label class="govuk-label govuk-checkboxes__label" for="@id">
             <span class="govuk-visually-hidden">Appoint</span>
-            @Formatters.capitalisedAddress(address)
+            <span name="address-text">@Formatters.capitalisedAddress(address)</span>
```
darrell-plant, line 58: why 'name'?

darrell-plant, line 58: I see :)

PeterPerhac: I don't. Why does the span have a name?

PeterPerhac: `name` is not a valid attribute of the `<span>` tag

darrell-plant: I deleted my original comment which was "Why 'name'". At first, I wondered why not have an `id` attribute, but then I had a quick look at how this code was used - i.e. within a loop. So, I thought maybe a non-unique attribute made sense (so I added 'I see'). But, tbh, I am not sure. My guess was that this attribute has been specifically requested as a means of pulling out all addresses in a selenium test. 

But yes, `name` is an odd choice. I think it is valid to have arbitrary attributes in html tags though. So, I think this is valid.

PeterPerhac: you can have arbitrary `data-` attributes. So `data-name` would be ok, though I don't see why it would be needed. If the element needs an ID, then let's give it an ID. 

https://www.w3.org/TR/REC-html40/index/attributes.html

```

name | BUTTON, TEXTAREA | CDATA | #IMPLIED |   |   |  
name | APPLET | CDATA | #IMPLIED | D | L | allows applets to find each other
name | SELECT | CDATA | #IMPLIED |   |   | field name
name | FORM | CDATA | #IMPLIED |   |   | name of form for scripting
name | FRAME, IFRAME | CDATA | #IMPLIED |   | F | name of frame for targetting
name | IMG | CDATA | #IMPLIED |   |   | name of image for scripting
name | A | CDATA | #IMPLIED |   |   | named link end
name | INPUT, OBJECT | CDATA | #IMPLIED |   |   | submit as part of form
name | MAP | CDATA | #REQUIRED |   |   | for reference by usemap
name | PARAM | CDATA | #REQUIRED |   |   | property name
name | META | NAME | #IMPLIED |   |   | metainformation name

```

darrell-plant: @WillBlakeney ?

WillBlakeney: The tests need to be able to pull out **all** addresses on the page, so there needs to be a generic locator to do this. At the moment the <label> has the hidden text "Appoint" within it, so when pulling out the addresses I get a list of addresses with the word "Appoint" before them. I have a way around this by mapping them and replacing the word, but I thought it would be neater to have one generic locator so I don't have to do that.

darrell-plant: That's what I assumed. Technically, `data-name`, or better still, `data-address` would be more correct.

darrell-plant: @dan.smith1 pls use `data-address` instead of `name`

app/views/propertyrepresentation/revokeAgentProperties.scala.html
```diff
@@ -55,7 +55,7 @@
         <input class="govuk-checkboxes__input" id="@id" name="linkIds[]" type="checkbox" value="@value">
         <label class="govuk-label govuk-checkboxes__label" for="@id">
             <span class="govuk-visually-hidden">Appoint</span>
-            @Formatters.capitalisedAddress(address)
+            <span data-address="address-text">@Formatters.capitalisedAddress(address)</span>
```
PeterPerhac, line 58: it doesn't need to have a value. `<span data-address>...</span>` can be used to label the address spans

DanSmith98: I see - I hadn't come across these data- attributes before. I'll update now

PeterPerhac:
👍  LGTPP

darrell-plant:
LGTDz 👍

PR #1039, opened 2021-11-29
VTCCA-4395 updated appoint an agent to one or more properties page la…

DanSmith98:
…yout


habeeb01:
LGTM 👍  (i actually thought I already gave you a thumb)

PR #1041, opened 2021-12-03
VTCCA-4395 dropdown not remembering selected agent fix

DanSmith98:
https://jira.tools.tax.service.gov.uk/browse/VTCCA-4395

### DIFF
--- a/app/views/propertyrepresentation/appoint/appointAgentProperties.scala.html
+++ b/app/views/propertyrepresentation/appoint/appointAgentProperties.scala.html
@@ -231,15 +231,15 @@
                             ),
                             items = SelectItem(
                                         value = params.agent,
-                                        text = if (params.agent.isDefined) {
-                                            params.agent.get
-                                        } else if (organisationAgents.resultCount > 1 && agentAlreadyAppointed) {
-                                            messages("propertyRepresentation.appointAgentProperties.dropdown.multiple", organisationAgents.resultCount - 1)
-                                        } else if (organisationAgents.resultCount > 1 && !agentAlreadyAppointed) {
-                                            messages("propertyRepresentation.appointAgentProperties.dropdown.multiple", organisationAgents.resultCount)
-                                        } else messages("propertyRepresentation.appointAgentProperties.dropdown.single")
+                                        text = params.agent.getOrElse(
+                                            if (organisationAgents.resultCount > 1 && agentAlreadyAppointed) {
+                                                messages("propertyRepresentation.appointAgentProperties.dropdown.multiple", organisationAgents.resultCount - 1)
+                                            } else if (organisationAgents.resultCount > 1 && !agentAlreadyAppointed) {
+                                                messages("propertyRepresentation.appointAgentProperties.dropdown.multiple", organisationAgents.resultCount)
+                                            } else messages("propertyRepresentation.appointAgentProperties.dropdown.single")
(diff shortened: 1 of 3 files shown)


app/views/propertyrepresentation/appoint/appointAgentProperties.scala.html
```diff
@@ -230,17 +230,20 @@ <h2 class="govuk-heading-m">@messages("propertyRepresentation.search.filter")</h
...
                                         } else if (organisationAgents.resultCount > 1 && !agentAlreadyAppointed) {
                                             messages("propertyRepresentation.appointAgentProperties.dropdown.multiple", organisationAgents.resultCount)
                                         } else messages("propertyRepresentation.appointAgentProperties.dropdown.single")
-                            ) +: organisationAgents.agents.collect{case agent if agent.organisationId != model.agentGroup.id =>
-                              SelectItem(
-                                  value = Some(agent.name),
-                                  text = agent.name
-                              )
+                            ) +: organisationAgents.agents.collect {
+                                case agent if (agent.organisationId != model.agentGroup.id) && (agent.name != params.agent.getOrElse("")) =>
```
darrell-plant, line 242:
```
agent.name != params.agent.getOrElse("")
```
could be something like
```
!params.agent.contains(agent.name)
```

darrell-plant:
LGTDz 👍

app/views/propertyrepresentation/appoint/appointAgentProperties.scala.html
```diff
@@ -230,17 +230,20 @@ <h2 class="govuk-heading-m">@messages("propertyRepresentation.search.filter")</h
                                 content = Text(messages("label.propertyRepresentation.appointedAgents"))
                             ),
                             items = SelectItem(
-                                        value = None,
-                                        text = if(organisationAgents.resultCount > 1 && agentAlreadyAppointed) {
+                                        value = params.agent,
+                                        text = if (params.agent.isDefined) {
```
PeterPerhac, line 234: don't use `isDefined` followed by `.get` on Options
Basically what you want to do here is `getOrElse`

```scala
//extract the 3 ELSE clauses to a lazy val (lazy as they might not need to be computed at all
lazy val emptyAgentText = if (organisationAgents.resultCount > 1 && agentAlreadyAppointed) {
      messages("propertyRepresentation.appointAgentProperties.dropdown.multiple", organisationAgents.resultCount - 1)
  } else if (organisationAgents.resultCount > 1 && !agentAlreadyAppointed) {
      messages("propertyRepresentation.appointAgentProperties.dropdown.multiple", organisationAgents.resultCount)
  } else messages("propertyRepresentation.appointAgentProperties.dropdown.single")

// ...

// then use getOrElse
text = params.agent.getOrElse(emptyAgentText)
```

PeterPerhac:
do you fancy writing a UNIT test for this change? So in the future we don't rely on testers knowing to check for this? This sort of thing could get checked during the SBT build

app/views/propertyrepresentation/appoint/appointAgentProperties.scala.html
DanSmith98: Looks like a much more elegant way of handling this, I'll change it now, thanks.

PeterPerhac:
👍  looks good to me

PR #1042, opened 2021-12-08
VTCCA-4540 - Added in feature flag for new registration journey

WitchellJack:
https://jira.tools.tax.service.gov.uk/browse/VTCCA-4540


darrell-plant:
LGTDz 👍

PR #1044, opened 2021-12-13
VTCCA-4511: updated error focus for date

zahiriqbal:


app/views/propertyLinking/ownershipToProperty.scala.html
```diff
@@ -55,7 +55,7 @@
 
 @getEnrichedErrorLinks = @{
     toCustomDateFormErrors.toSet.map{ error: FormError =>
-        ErrorLink(href = Some(s"#${error.key}"),
+        ErrorLink(href = Some(s"#${error.key}.day"),
```
darrell-plant, line 58: why `day`? (what about month and year?)

zahiriqbal: @darrell.plant I spoke to @will.blakeney and he said only Day field should be focused, as the error summary message is only one and doesn't break down into day, month and year.

darrell-plant: yep, I was just puzzled, but makes sense

darrell-plant: and before adding `day` it didn't work - got it now :)

app/views/propertyLinking/ownershipToProperty.scala.html
```diff
@@ -126,7 +126,7 @@
                 item =>
                     InputItem(
                         id = s"toDate.$item",
-                        classes = s"govuk-input--width-${if(item == "year") "4" else "2"}",
+                        classes = s"govuk-input--width-${if(item == "year") "4" else "2"}${if(form(s"toDate.$item").hasErrors || form.errors.exists(_.key.isEmpty)) " govuk-input--error" else ""}",
```
darrell-plant, line 129: this 
```
s"govuk-input--width-${if(item == "year") "4" else "2"}${if(form(s"toDate.$item").hasErrors || form.errors.exists(_.key.isEmpty)) " govuk-input--error" else ""}"
```
is repeated, so could be extracted to method, but not a biggie

PeterPerhac:
👍  looks good to me

PR #1055, opened 2022-01-25
VTCCA-4483 Updated Confirmation pages

reddy-yattapu:
https://jira.tools.tax.service.gov.uk/browse/VTCCA-4483

### DIFF
--- a/app/controllers/IdentityVerification.scala
+++ b/app/controllers/IdentityVerification.scala
@@ -18,6 +18,7 @@ package controllers
 
 import actions.registration.GgAuthenticatedAction
 import config.ApplicationConfig
+import controllers.registration.routes
 import models.identityVerificationProxy.IvResult
 import models.registration._
 import play.api.Logging
@@ -26,8 +27,8 @@ import play.api.mvc._
 import repositories.SessionRepo
 import services.iv.IdentityVerificationService
 import uk.gov.hmrc.propertylinking.errorhandler.CustomErrorHandler
-
 import javax.inject.{Inject, Named}
+
 import scala.concurrent.{ExecutionContext, Future}
 
 class IdentityVerification @Inject()(
@@ -73,7 +74,9 @@ class IdentityVerification @Inject()(
         case true =>
           identityVerificationService.continue(journeyId, request.userDetails).map {
             case Some(RegistrationSuccess(personId)) =>
-              Redirect(controllers.registration.routes.RegistrationController.success(personId))
+              if (config.newRegistrationJourneyEnabled)
+                Redirect(controllers.registration.routes.RegistrationController.confirmation(personId))
+              else Redirect(controllers.registration.routes.RegistrationController.success(personId))
             case _ => InternalServerError(errorHandler.internalServerErrorTemplate)
           }
         case false => Future.successful(Unauthorized(errorHandler.internalServerErrorTemplate))
--- a/app/views/createAccount/confirmation.scala.html
+++ b/app/views/createAccount/confirmation.scala.html
@@ -1,0 +1,58 @@
+@*
+ * Copyright 2021 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import helpers._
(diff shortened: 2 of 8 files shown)


test/controllers/RegistrationControllerSpec.scala
```diff
@@ -238,7 +240,26 @@ class RegistrationControllerSpec extends VoaPropertyLinkingSpec with MockitoSuga
...
+
+    val res =
+      testRegistrationController(userDetails(confidenceLevel = ConfidenceLevel.L200)).confirmation(100)(fakeRequest)
+    status(res) shouldBe OK
+    val html = HtmlPage(res)
+    //Page title
+    html.shouldContainText("Registration successful")
+    html.shouldContainText("Your VOA personal ID number")
+    //Page should contains VOA Person ID value 100
+    html.shouldContainText("100")
```
PeterPerhac, line 258: I think this test is not quite good enough. There could be a variety of other things on the page that will include the number 100. e.g. "You have 100 properties", or check reference "CHK1002345" or just some sort of css class could be "padding-top-100" or something like that... just checking if the HTML contains 100 (or 300) is not good enough test. Use a bit more text maybe to check it's in the page?

conf/messages
```diff
@@ -280,6 +281,16 @@ createAccount.group.details.header.multipleAddress=There’s more than one addre
 createAccount.group.details.content.multipleAddress=You should enter the address that you want any business correspondence to be sent to.
 createAccount.business.email.help=This should be your email address and not your agent’s. The email address should be monitored.
 
+registration.successful=Registration successful
+voa.personal.id.number=Your VOA personal ID number
+agent.code=Your Agent code
+email.sent.to=We have sent these details to <strong>{0}</strong>
+next.time.sign.in=Next time you sign in to the service, use your Government Gateway details.
+reset.password=If you need to reset your password, you'll need to provide your VOA personal ID number.
+give.agent.code.to.clients=Give your Agent code to your client so they can appoint you to act for them.
```
PeterPerhac, line 290: the message keys are not what we usually do, i think... in `give.agent.code.to.clients`, the text could be later changed to say something totally different, e.g. `give.agent.code.to.clients=Contact the VOA for a free valuation` and then they the key will make no sense. I think we used hierarchy of where the text sits on the page, like `pageName.sectionName.p1` ... I am not too bothered about this, but I wonder what others think

darrell-plant: let's get it right 🙂
key must start with `pageName` at least - after that it gets tricky, but the key should be based on function or page-structure (as pete said), not on content of the message

app/controllers/IdentityVerification.scala
```diff
@@ -73,7 +74,9 @@ class IdentityVerification @Inject()(
         case true =>
           identityVerificationService.continue(journeyId, request.userDetails).map {
             case Some(RegistrationSuccess(personId)) =>
-              Redirect(controllers.registration.routes.RegistrationController.success(personId))
+              if (config.newRegistrationJourneyEnabled)
+                Redirect(controllers.registration.routes.RegistrationController.confirmation(personId))
+              else Redirect(controllers.registration.routes.RegistrationController.success(personId))
```
darrell-plant, line 79: can we use an import to avoid fully-qualified names

app/views/createAccount/confirmation.scala.html
```diff
@@ -0,0 +1,58 @@
+@*
+ * Copyright 2021 HM Revenue & Customs
```
darrell-plant, line 2: It's 2022 🙂

app/controllers/IdentityVerification.scala
reddy-yattapu: Controller are in mutiple packages and this IdentityVerification.scala using reference to different controllers routes so needed to have qualified names

conf/messages
reddy-yattapu: keys updated

test/controllers/RegistrationControllerSpec.scala
reddy-yattapu: I have updated the test

app/views/createAccount/confirmation.scala.html
reddy-yattapu: I have updated the year

app/controllers/IdentityVerification.scala
darrell-plant: I think you can remove the unused import on line 21 and do 
```
if (config.newRegistrationJourneyEnabled)
                Redirect(registration.routes.RegistrationController.confirmation(personId))
              else Redirect(registration.routes.RegistrationController.success(personId))
```
that would be a little better

and same on line 64

I find `routes` quite tricky tbh

app/views/createAccount/confirmation.scala.html
darrell-plant: I was joking, but thanks. All our twirl template headers are 2021 across all services. Something we will probably fix somehow, sometime.

darrell-plant:
LGTDz 👍 (with one minor comment)

PeterPerhac:
👍  looks good but formatting is off, pr builder fails

PR #1058, opened 2022-02-01
VTCCA-4593 update request a dvr confirmation page

michjanowicz:
https://jira.tools.tax.service.gov.uk/browse/VTCCA-4593

### DIFF
--- a/app/views/dvr/requestedDetailedValuation.scala.html
+++ b/app/views/dvr/requestedDetailedValuation.scala.html
@@ -16,25 +16,55 @@
 
 @import config.ApplicationConfig
 @import utils.Formatters
+@import views.html.includes.assessmentHeading
 
 @this(mainLayout: views.html.mainLayout)
 
-@(submissionId: String, address: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
+@(submissionId: String, address: String,  localAuthorityRef: String, clientDetails: Option[ClientDetails] = None)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
 
-@mainLayout(pageTitle=Messages("detailedValuationRequested.page.title")) {
+
+@managePropertiesUrl = @{
+    clientDetails match {
+        case Some(client) =>
+            config.dashboardUrl(
+                s"selected-client-properties?clientOrganisationId=${client.organisationId}&clientName=${client.organisationName}")
+        case _        => config.dashboardUrl("your-properties")
+    }
+}
+
+@linkSubmitted = @{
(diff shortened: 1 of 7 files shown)


app/views/includes/assessmentHeading.scala.html
```diff
@@ -6,12 +6,12 @@
 <div id="assessment-heading">
     <ul id="assessment-heading" class="govuk-list">
         <li id="local-authority-reference">
-            <span class="govuk-body-s govuk-!-font-weight-bold">@messages("assessments.localAuthorityRef")</span>
-            <span class="govuk-body-s" id="local-authority-reference-code">@localAuthorityReference</span>
+            <span class="govuk-body-m govuk-!-font-weight-bold">@messages("assessments.localAuthorityRef")</span>
```
darrell-plant, line 9: This change is outside the scope of the Jira and will affect all pages. Are you bringing `assessmentHeading` inline with other services here?

darrell-plant:
LGTDz 👍

app/views/includes/assessmentHeading.scala.html
michjanowicz: I made font size bigger, following prototype. I know is affecting all the views in PL, however, recently  I did https://jira.tools.tax.service.gov.uk/browse/VTCCA-4471, where the task was to increase font size across the whole check journey.

app/views/dvr/requestedDetailedValuation.scala.html
```diff
@@ -16,25 +16,55 @@
...
-@(submissionId: String, address: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
+@(submissionId: String, address: String,  localAuthorityRef: String, clientDetails: Option[ClientDetails] = None)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
 
-@mainLayout(pageTitle=Messages("detailedValuationRequested.page.title")) {
+
+@managePropertiesUrl = @{
+    clientDetails match {
+        case Some(client) =>
+            config.dashboardUrl(
+                s"selected-client-properties?clientOrganisationId=${client.organisationId}&clientName=${client.organisationName}")
```
PeterPerhac, line 30: string interpolation like this isn't safe for URLs but I think we do this in many other places. I think it will be good to have a test of what happens when a client org name is "foo&clientName=bar" and review the current situation in codebase, just how widely used is string interpolation and fix it - URL-encoded strings or use the `uri-template` library (out of scope for this ticket)

PeterPerhac:
👍  looks ok to me

PR #1066, opened 2022-02-17
VTCCA-4617-connection-to-the-property-page-layout-updated

michjanowicz:
https://jira.tools.tax.service.gov.uk/browse/VTCCA-4617


darrell-plant:
LGTDZ 👍

PeterPerhac:
👍  looks good to me

one minor suggestion: when naming git branches, I would say `VTCCA-4617-connection-to-the-property-page-layout-updated` is a bit too long. Using just VTCCA-4617 is too short and not enough to be able to tell by just looking what kind of branch it is (without looking it up in jira) but adding just a little bit of context to the branch name is enough, so for example `VTCCA-4617-property-page-layout` is enough.

PR #1067, opened 2022-02-18
VTCCA-4617 connection to the property page layout update

michjanowicz:
This PR has been approved earlier on: https://github.com/hmrc/voa-property-linking-frontend/pull/1066 Changes after: https://hmrcdigital.slack.com/archives/G3QPKL2V9/p1645140455794799 Ticket: https://jira.tools.tax.service.gov.uk/browse/VTCCA-4617


conf/messages
```diff
@@ -778,14 +778,14 @@ toDate.year.label=Year
...
+declareCapacity.details.ownerOrOccupier=I own or occupy a part of this property
+declareCapacity.details.ownerOrOccupier.text=After the Valuation Office Agency have approved your connection, you may want to ask them to split this property to get separate valuations for each part.
+declareCapacity.details.notOwner=I do not own the property but I sublet it to someone else
+declareCapacity.details.notOwner.text=<p>If you sublet the property to another person or business, select ‘Owner’ and enter the dates of the sublet period.</p><p>If you only sublet part of the property, you may want to ask the Valuation Office Agency to split this property to get separate valuations for each part. You can do this after they have approved your connection to the property.</p>
+declareCapacity.details.ownerOrOccupier.client=My client owns or occupies part of this property
+declareCapacity.details.ownerOrOccupier.text.client=<p>You can still add this property on behalf of your client, as if they owned or occupied the whole property.</p><p>After we’ve approved your client’s connection to the property, you can ask us to split the property to get separate valuations for each part.</p>
```
habeeb01, line 786: are you missing `class="govuk-body"` on your p tags

michjanowicz: This class should be added directly in messages? This <p>'s are wrapped in div "govuk-details_text" as a part of govukDetails component.

michjanowicz: This p's are wrapped in div "govuk-details_text" as a part of govukDetails component.*

habeeb01: i personally don't like html in messages file. I know that there are lots of others here as well (may be we'll handle them as a separate task). Are you saying that the class is not needed because the styles will be inherited from "govuk-details_text"? So long as you've tested this and that is the case then im happy 👍

michjanowicz: Yes. I did test it. This solution was present in code before. I just swap text in <p>. Thanks for comment.

WitchellJack:
LGTM 👍🏻

PR #1068, opened 2022-02-18
VTCCA-4720 added aria-lable

zahiriqbal:
https://jira.tools.tax.service.gov.uk/browse/VTCCA-4720

### DIFF
--- a/app/views/dashboard/assessmentDetails.scala.html
+++ b/app/views/dashboard/assessmentDetails.scala.html
@@ -1,18 +1,18 @@
 @*
- * Copyright 2021 HM Revenue & Customs
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- *@
+* Copyright 2021 HM Revenue & Customs
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*@
 
 @import actions.requests.AuthenticatedRequest
 @import actions.assessments.request.AssessmentsPageSessionRequest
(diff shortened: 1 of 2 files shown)


darrell-plant:
code duplication needs fixing 
then hanging " to " needs fixing
logic to determine Current|Draft|Historic is not right
also, odd formattting

app/views/dashboard/assessmentDetails.scala.html
```diff
@@ -33,10 +32,36 @@
...
+        (assessment.currentFromDate, assessment.currentToDate) match {
+            case (Some(_), None)     => Some(Messages("assessments.enddate.present"))
+            case (Some(_), Some(to)) => Some(Formatters.formatDate(to))
+            case _                   => None
+        }
+    } else None
+}
+
+@ariaLabel(assessment: ApiAssessment, listType: String, isDraft: Boolean) = @{
+    val listTypeText = listType.takeWhile(_ != 'A').capitalize
```
PeterPerhac, line 54: what does this line do?

PeterPerhac: `listType.takeWhile(_ != 'A').capitalize`

app/views/dashboard/assessmentDetails.scala.html
```diff
@@ -33,10 +32,36 @@
...
+    val endText = endDate(assessment, isDraft).map(ed => s"to $ed").getOrElse("").trim
+
+    s"Select this valuation - $listTypeText $fromText $endText"
+}
+
+@table(list: Seq[(String, ApiAssessment)], listType: String, isDraft: Boolean) = {
+    <div id="viewAssessmentRadioGroup_@listType" class="form-group">
         <div class="table-wrapper-no-pad">
-            <table id="@tableId" class="margin-bottom-10 margin-top-10">
+            <table id="@listType" class="margin-bottom-10 margin-top-10">
```
PeterPerhac, line 64: what is `listType` ? is it safe to use this String as an HTML ID attribute?

app/views/dashboard/assessmentDetails.scala.html
```diff
@@ -120,13 +140,13 @@
 <div id="details-content-valuations" class="govuk-body">
 
     @if(model.currentAssessments.nonEmpty){
-    <h2 class="govuk-heading-m">@messages("assessments.current.title")</h2>
-        @table(model.currentAssessments, "currentAssessments")
+        <h2 class="govuk-heading-m">@messages("assessments.current.title")</h2>
+        @table(model.currentAssessments, listType = "currentAssessments", isDraft = false)
```
PeterPerhac, line 144: I don't understand where does this "currentAssessments" come from? That's not a list type...

app/views/dashboard/assessmentDetails.scala.html
```diff
@@ -57,53 +82,48 @@
...
                     @list.map {  case (url, assessment) =>
-                        <tr class="govuk-table__row">
-                            <td class="govuk-table__cell">
-                                @assessment.effectiveDate.map(Formatters.formatDate).getOrElse("")
-                            </td>
+                    <tr class="govuk-table__row">
+
+                        <td class="govuk-table__cell">@effectiveDate(assessment)</td>
+
+                        @assessment.rateableValue match {
```
PeterPerhac, line 89: un thus template I see the same kind of problem repeatedly:

```
 @assessment.rateableValue match {
                            case Some(rv) => {
                                <td class="govuk-table__cell">£@rv</td>
                            }
                            case None => {
                                <td class="govuk-table__cell">@messages("assessments.na")</td>
                            }
                        }
```

regardless of whether the value is Some or None, you'd always want to output the  `<td class="govuk-table__cell">` and the `</td>`

Why repeat the element?

this sort of code is what you want I think:

```
<td class="govuk-table__cell">
  @assessment.rateableValue.fold(messages("assessments.na"))( rv => s"£$rv" )
</td>
```

app/views/dashboard/assessmentDetails.scala.html
darrell-plant: yes - and it hasn't changed - just the name.

app/views/dashboard/assessmentDetails.scala.html
darrell-plant: Splits the input "currentAssessments" (for example) to "Current"

app/views/dashboard/assessmentDetails.scala.html
darrell-plant: explained offline

app/views/dashboard/assessmentDetails.scala.html
darrell-plant: true - it's existing code, but I will tidy that up

app/views/dashboard/assessmentDetails.scala.html
darrell-plant: agreed approach

zahiriqbal:
👍

PR #1081, opened 2022-02-25
VTCCA-4758 fix broken empty checks tab on dvr screen

PeterPerhac:
https://jira.tools.tax.service.gov.uk/browse/VTCCA-4758


reddy-yattapu:
👍

PR #1084, opened 2022-03-04
VTCCA-4641 other evidence page

habeeb01:
https://jira.tools.tax.service.gov.uk/browse/VTCCA-4641

### DIFF
--- a/app/assets/javascripts/src/updateEvidenceType.js
+++ b/app/assets/javascripts/src/updateEvidenceType.js
@@ -1,0 +1,74 @@
+(function() {
+
+    'use strict';
+
+    var root = this, $ = root.jQuery;
+
+    if (typeof VOA === 'undefined') {
+        root.VOA = {};
+    }
+
+    var UpdateEvidenceType = function (){
+
+        var errorMessages = '<div id="error-summary" class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">'+
+                '<h2 class="govuk-error-summary__title" id="error-summary-heading">There is a problem</h2>'+
+                '<ul class="govuk-list govuk-error-summary__list"><li></li></ul></div>';
+
+
+        var options = "input[value='lease'],input[value='license'],input[value='serviceCharge'],input[value='stampDutyLandTaxForm'],input[value='waterRateDemand'],input[value='otherUtilityBill'],input[value='landRegistryTitle'],input[value='unableToProvide']";
+
+        $(options).click(function() {
+          submitEvidenceType(this.value);
+        });
+
+       function submitEvidenceType(evidenceType){
+                    if (evidenceType !== "") {
+                        var csrfToken = $("#uploadForm input[name='csrfToken']").val();
+                        clearErrors("evidenceType");
+                        $.ajax({
+                            url: $("#updateEvidenceTypeURL").text(),
+                            method: "POST",
+                            contentType: "application/json",
+                            data: JSON.stringify({
+                                "evidenceType": evidenceType,
+                                "csrfToken": csrfToken
+                            }),
+                            headers: {
+                                'Csrf-Token': csrfToken
+                            },
+                            crossDomain: false,
+                            cache: false
+                        }).fail(function (jqXHR) {
+                            addError(jqXHR.responseText, "evidenceType");
+                        }).done(function (data, statusText, resObject) {
+                            $('#message-warning').addClass('govuk-visually-hidden');
+                        });
+                    }
+               }
+
+        var errorTitlePrefix = 'Error: ';
+        var title = $(document).prop('title');
+        function addError(message, evidenceTypeId){
+            if(title.startsWith(errorTitlePrefix)){
+                $(document).prop('title', title);
+            } else {
+                $(document).prop('title', errorTitlePrefix + title);
+            }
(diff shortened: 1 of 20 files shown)


app/views/propertyLinking/uploadEvidence.scala.html
```diff
@@ -59,108 +63,135 @@ <h2 class="govuk-error-summary__title" id="error-summary-heading">@messages("err
...
                     ))
                 }
 
+                <div class="message-warning govuk-visually-hidden" id="message-warning"><p class="govuk-body">@messages("fileupload.pleaseWait")</p></div>
+                <table class="govuk-table">
+                    <tbody class="govuk-table__body">
+                    @uploadedFiles.map { case (fileReference, upload) =>
+                    <tr class="govuk-table__row">
+                        <td class="govuk-table__cell">@{upload.fileMetadata.toDisplayFileName}</td>
+                        <td class="govuk-table__cell">@*(@upload.size.formatted)*@</td>
```
PeterPerhac, line 166: File size is commented out ?

app/assets/javascripts/src/updateEvidenceType.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    var UpdateEvidenceType = function (){
+
+        var errorMessages = '<div id="error-summary" class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">'+
+                '<h2 class="govuk-error-summary__title" id="error-summary-heading">There is a problem</h2>'+
+                '<ul class="govuk-list govuk-error-summary__list"><li></li></ul></div>';
+
+
+        var options = "input[value='lease'],input[value='license'],input[value='serviceCharge'],input[value='stampDutyLandTaxForm'],input[value='waterRateDemand'],input[value='otherUtilityBill'],input[value='landRegistryTitle'],input[value='unableToProvide']";
```
PeterPerhac, line 18: this is too brittle. If any of the values is changed in the template, then one of the inputs will not be picked up by this javascript and it will be hard to spot the bug. We can pick all the inputs within a `fieldset` with a selector.

app/models/FileInfo.scala
```diff
@@ -18,7 +18,7 @@ package models
 
 import play.api.libs.json._
 
-case class FileInfo(name: String, evidenceType: Option[EvidenceType])
+case class FileInfo(name: Option[String], evidenceType: Option[EvidenceType])
```
darrell-plant, line 21: Is it necessary to make `name` an Option. I'm not sure that name should/could ever be None.

app/assets/javascripts/src/updateOtherEvidencePage.js
```diff
@@ -0,0 +1,35 @@
...
+        root.VOA = {};
+    }
+
+    var UpdateOtherEvidencePage = function (){
+
+        $("#continue-button").click(function() {
+          $('#continue').click();
+        });
+
+        var options = "input[value='lease'],input[value='license'],input[value='serviceCharge'],input[value='stampDutyLandTaxForm'],input[value='waterRateDemand'],input[value='otherUtilityBill'],input[value='landRegistryTitle']";
```
PeterPerhac, line 17: also here, listing inputs by their value attribute is brittle. We can list all options within a specific `fieldset`

something like this
```javascript
$('fieldset#fieldset-id input[type="checkbox"]')
```

app/views/propertyLinking/uploadEvidence.scala.html
habeeb01: 🦅 👁️  I hadn't noticed that but I have checked now and  it was previously like that (not sure why). I will see if it can be removed all together.

app/models/FileInfo.scala
habeeb01: it needs to be optional. We can have a huddle to discuss why if you'd like...in short, we now need to save the evidence type before the file is uploaded.

app/assets/javascripts/src/updateEvidenceType.js
```diff
@@ -0,0 +1,74 @@
...
+
+        $(options).click(function() {
+          submitEvidenceType(this.value);
+        });
```
darrell-plant, line 22: Does this add the click event to all the inputs? If so, maybe there's a more concise selector that doesn't require knowing all the values? Maybe a class - or `input[type="radio"]`?

app/assets/javascripts/src/updateOtherEvidencePage.js
darrell-plant, line 17: same here (concise css selector)

app/views/propertyLinking/declaration.scala.html
```diff
@@ -51,7 +51,7 @@
 
 @propertyStillOccupied = @{request.ses.propertyOccupancy.exists(_.stillOccupied)}
 
-@strippedFileName = @{request.ses.uploadEvidenceData.fileInfo.map(_.name.replaceAll("^([^-]+)-", ""))}
+@strippedFileName = @{request.ses.uploadEvidenceData.fileInfo.map(_.name.fold("")(_.replaceAll("^([^-]+)-", "")))}
```
darrell-plant, line 54: I guess this is because filename can now be None? But it probably should never be None.

darrell-plant:
LGTDz 👍

PeterPerhac:
👍  looks good to PP

PR #1092, opened 2022-03-18
VTCCA-4798 comma format assessments rateable values

DanSmith98:


app/utils/Formatters.scala
```diff
@@ -58,4 +59,18 @@ object Formatters {
...
     value match { case Some(paramValue) if paramValue != "" => s"&$name=${paramValue.toUpperCase.trim}"; case _ => "" }
 
+  def formatCurrency[A: Numeric](total: A): String = {
+    val currencyFormatter = getCurrencyInstance(java.util.Locale.UK)
+    currencyFormatter.setRoundingMode(java.math.RoundingMode.DOWN)
+
+    currencyFormatter.format(total)
+  }
+
+  def formatCurrencyRoundedToPounds[A: Numeric](total: A): String = {
```
PeterPerhac, line 69: could we add a `maximumFractionDigits:Option[Int] = None` parameter to `formatCurrency` and use it like so:

```scala
maximumFractionDigits.foreach(digits => currencyFormatter.setMaximumFractionDigits(digits))
```

then the `formatCurrencyRoundedToPounds` method would pass-through to it, like so `formatCurrency(total, Some(0))`

PeterPerhac:
This looks good, I just find that the two currency formatting methods are largely similar, with only 1 additional line in one of them. I would like you to express the second one in terms of the first one.

app/utils/Formatters.scala
DanSmith98: I like it, updating now

PeterPerhac:
👍  looks good to me

habeeb01:
LGTM 👍

PR #1093, opened 2022-03-21
VTCCA-4574 ensure csv file cannot be uploaded

habeeb01:


darrell-plant:
LGTDz 👍

PeterPerhac:
only 1 problem with this PR, I didn't get a chance to review it yesterday - and the problem is in all three frontends

as soon as the user has a csv file like so "somefile.CSV" or "somefile.Csv" it will not stop him. Need to make a case-insensitive comparison. or `.toLowerCase()` the extension first, before comparing with "csv"

PR #1101, opened 2022-04-20
VTCCA-0 set newRegistrationJourneyEnabled to true

darrell-plant:
As the flag is set to true in QA in app-config, we should also set to true in local


habeeb01:
LGTM 👍

PR #1117, opened 2022-06-01
VTCCA_4907 error message bugfix

darrell-plant:
text changes
error summary agreed fix
started to address the misuse of `class="govuk-visually-hidden"`

### DIFF
--- a/app/views/propertyLinking/uploadEvidence.scala.html
+++ b/app/views/propertyLinking/uploadEvidence.scala.html
@@ -21,8 +21,10 @@
 @import models._
 @import views.html.includes.assessmentHeading
 @import actions.propertylinking.requests.LinkingSessionRequest
+@import _root_.utils.Formatters._
 
 @this(govukErrorSummary : GovukErrorSummary, govukRadios: GovukRadios, govukButton : GovukButton, mainLayout: views.html.mainLayout, formWithCSRF: FormWithCSRF, govukWarningText: GovukWarningText)
+
 
 @(
         submissionId: String,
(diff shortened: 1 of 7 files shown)


app/views/propertyLinking/uploadEvidence.scala.html
```diff
@@ -125,9 +129,9 @@ <h1 class="govuk-heading-l">@messages("uploadOtherEvidence.title")</h1>
         ))
     }
 
-    <div id="file-upload-form" class="govuk-visually-hidden">
+    <div id="file-upload-form" class="govuk-!-display-none">
         @formWithCSRF(action = UploadController.initiate(EvidenceChoices.OTHER), 'enctype -> "multipart/form-data", 'id -> "uploadForm") {
-            <div class="govuk-visually-hidden" id="initiateFields"></div>
+            <div id="initiateFields" class="govuk-!-display-none"></div>
```
reddy-yattapu, line 134: I think we all following html Ids pattern ='abc-edf-xyz' rather camel case

reddy-yattapu:
👍 with very small comment

app/views/propertyLinking/uploadEvidence.scala.html
darrell-plant: That `id` is already there and I didn't want to change it as it may be referenced in javascript.

habeeb01:
LGTM 👍

app/views/propertyLinking/uploadEvidence.scala.html
reddy-yattapu: no problem , I got it

PR #1118, opened 2022-06-08
VTCCA-4906 use cache id for back link when provided

habeeb01:


darrell-plant:
LGTDz 👍

app/uk/gov/hmrc/propertylinking/controllers/valuations/ValuationsController.scala
```diff
@@ -102,7 +106,12 @@ class ValuationsController @Inject()(
         .flatMap {
           case Some(EmptyAssessments()) | None => Future.successful(notFound)
           case Some(assessments) =>
-            if (owner) Future.successful(okResponse(assessments, config.dashboardUrl("your-properties")))
+            if (owner)
+              Future.successful(
+                okResponse(
+                  assessments,
+                  s"${config.dashboardUrl("return-to-your-properties")}${request.sessionData.returnSearchCacheId.fold(
+                    "")(id => s"?mayBeSessionId=$id")}"))
```
PeterPerhac, line 114: I don't like the parameter name `mayBeSessionId`

habeeb01: (I'm not crazy about the name either but) changing `mayBeSessionId` at this point would mean updating 2 other services (as well as this one). This ticket is already blocking another on the board, no need to delay much longer. Please feel free to suggest a better name but it will be done as a separate piece of work

reddy-yattapu:
👍

app/uk/gov/hmrc/propertylinking/controllers/valuations/ValuationsController.scala
habeeb01: i decided to just bite the bullet and just do it. Im just running tests now. This PR will be updated very soon (will update the other services as well). I have gone with `cid`.

PR #1121, opened 2022-06-20
VTCCA-4982 update your details validation added

michjanowicz:


app/utils/PhoneNumberValidation.scala
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package utils
+
+import play.api.data.Forms.text
+
+object PhoneNumberValidation {
+
+  val phoneNumberRegex =
```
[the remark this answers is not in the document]
darrell-plant: for such a complex regex we should have unit tests that test all the possible forms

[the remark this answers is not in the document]
michjanowicz: pr has been updated

test/utils/PhoneNumberValidationSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+
+        if (isValid) {
+          res.hasErrors shouldBe false
+        } else {
+          res.errors(0).message shouldBe "error.phoneNumber.invalidFormat"
+        }
+      }
+
+    }
+    "catch phone validity" in new Setup {
```
darrell-plant, line 90: 'catch mandatory condition'

app/utils/PhoneNumberValidation.scala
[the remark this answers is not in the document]
darrell-plant: I think it would be useful to have a comment block above this regex to list the valid formats (one example of each - it looks like there are 7 alternatives - separated by `|`)

app/utils/PhoneNumberValidation.scala
```diff
@@ -20,6 +20,8 @@ import play.api.data.Forms.text
 
 object PhoneNumberValidation {
 
+  // Valid formats: 1234567890 | +44 1234567890 | +44 0123 456 789 | +441234567890 | 0123 456 7890 | 0123-456-7890 | +44 123-456-7890
```
darrell-plant, line 23: 👍

test/utils/PhoneNumberValidationSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+        } else {
+          res.errors(0).message shouldBe "error.phoneNumber.invalidLength"
+        }
+      }
+    }
+
+    "catch invalid format error" in new Setup {
+      val formats = Table(
+        ("phone number", "validity"),
+        ("phonenumber123456", false),
```
PeterPerhac, line 66: I don't think in this case we need the Table. All entires are expected to be invalid. So we could just run a list of invalid values through the validator and assert that forAll is valid = false

test/utils/PhoneNumberValidationSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+        ("9790 305 1234", false),
+        ("0890456782a", false),
+        ("+44-909-789-8973", false),
+        ("#447903067897", false),
+      )
+
+      TableDrivenPropertyChecks.forAll(formats) { (phone, isValid) =>
+        val res: Form[String] = form.bind(Map("phone" -> phone))
+
+        if (isValid) {
```
PeterPerhac, line 82: then this if-condition would not be necessary. as we'd expect all inputs to be invalid

PeterPerhac:
👍  looks good to me. The test thing I pointed out is minor nitpick. I think it would be nice if you removed the `if(isValid)` condition because all inputs are expected not to be valid. But it's not a biggie, I understand you used same style of tests across the whole file, so that's nice for consistency...

michjanowicz:
Is done. Thanks a lot for review and tips.

PR #1122, opened 2022-06-22
VTCCA-4989 updated the js so it only applies to the fileupload page n…

zahiriqbal:
…ot all pages


darrell-plant:
LGTDz 👍

DanSmith98:
LGTM :+1:

PR #1134, opened 2022-08-17
VTCCA-5019 Review unit tests in Property Linking journey

reddy-yattapu:


habeeb01:
there are still some test classes with the same issue: e.g. `ManageAgentControllerSpec` or `RepresentationControllerSpec` (there might be others). It would be good to get them all at the same time

habeeb01:
LGTM 👍

PR #1145, opened 2022-09-21
VTCCA-5185 Registration Complete Contact Details field validation cov…

zachhall1234:
…erage

### DIFF
--- a/test/controllers/RegistrationControllerSpec.scala
+++ b/test/controllers/RegistrationControllerSpec.scala
@@ -264,7 +264,36 @@ class RegistrationControllerSpec extends VoaPropertyLinkingSpec with MockitoSuga
 
   }
 
-  "Submitting an invalid individual form" should "return a bad request response" in {
+  "Submitting an individual form with invalid email, mobilePhone, phone, nino" should "return a bad request response" in {
+    val data = Map(
+      "phone"       -> Seq("01"),
+      "mobilePhone" -> Seq("11111222111"),
+      "nino"        -> Seq("0AQ"),
+      "email"       -> Seq("invalidEmail!!.com"))
+
+    val fakeRequest: FakeRequest[AnyContent] = FakeRequest().withBody(AnyContentAsFormUrlEncoded(data))
+
+    val res = testRegistrationController(userDetails()).submitIndividual()(fakeRequest)
+    status(res) shouldBe BAD_REQUEST
+    val html = HtmlPage(res)
+    html.shouldContainText("Enter a valid email address")
+    html.shouldContainText("Enter a valid National Insurance number")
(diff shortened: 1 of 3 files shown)


test/controllers/RegistrationControllerSpec.scala
```diff
@@ -264,7 +264,36 @@ class RegistrationControllerSpec extends VoaPropertyLinkingSpec with MockitoSuga
 
   }
 
-  "Submitting an invalid individual form" should "return a bad request response" in {
+  "Submitting an individual form with invalid email, mobilePhone, phone, nino" should "return a bad request response" in {
+    val data = Map(
```
zahiriqbal, line 268: You can move this to the FakeObjects class...

zahiriqbal, line 268: You can move this to the FakeObjects class...

test/controllers/RegistrationControllerSpec.scala
```diff
@@ -264,7 +264,36 @@ class RegistrationControllerSpec extends VoaPropertyLinkingSpec with MockitoSuga
...
+    status(res) shouldBe BAD_REQUEST
+    val html = HtmlPage(res)
+    html.shouldContainText("Enter a valid email address")
+    html.shouldContainText("Enter a valid National Insurance number")
+    html.shouldContainText("Telephone number must be between 11 and 20 characters")
+    html.shouldContainText("Enter a telephone number, like 01623 960 001 or +44 0808 157 0192")
+  }
+
+  "Submitting an individual form with non-matching emails" should "return a bad request response" in {
+    val data = Map("email" -> Seq("x@x.com"), "confirmedEmail" -> Seq("1@1.com"))
```
zahiriqbal, line 286: also this

test/controllers/RegistrationControllerSpec.scala
DanSmith98: I don't think this is likely to be used anywhere else, so I'm not sure putting in FakeObjects would be of benefit here.

test/controllers/RegistrationControllerSpec.scala
DanSmith98: I don't think this will be used anywhere else either

DanSmith98:
LGTM :+1:

test/controllers/RegistrationControllerSpec.scala
PeterPerhac: thanks @DanSmith98 you're right, these maps are useful only in this specific place. They belong in the test, not in the FakeObjects

PeterPerhac:
👍  looks good to me

PR #1156, opened 2022-10-11
VTCCA-5168 fix manual address link not showing, fix bugs resulting from fix, refactor address partial DI

DanSmith98:


habeeb01:
Dan to add a test to check that the "Enter address manually" link is displayed on load of the page

DanSmith98:
I've added some testing on the visibility of the manual entry link. It's not as comprehensive as I'd like, partially because of the javascript and partially because I can't run the 'update address' screens locally. I'll add a comment on the ticket to regression test those pages as well.

reddy-yattapu:
👍

habeeb01:
LGTM 👍

PR #1164, opened 2022-10-19
VTCCA-5208 fixed rateable value caption

habeeb01:


app/models/PropertyLink.scala
```diff
@@ -64,6 +64,9 @@ case class ApiAssessment(
 ) {
   def isWelsh: Boolean = billingAuthorityCode.exists(_.startsWith("6"))
   def isDraft: Boolean = listType == ListType.DRAFT
+
+  def isCurrent: Boolean = listType == ListType.CURRENT && currentFromDate.nonEmpty && currentToDate.isEmpty
+  def isPrevious: Boolean = !isDraft && !isCurrent
```
PeterPerhac, line 69: I don't see this `isPrevious` used anywhere else, it's a new def so should be referenced somewhere else in this PR but it isn't

PeterPerhac: also `isCurrent` appears to only be used on an assessment, not on a property link... so not sure if either of these two are used

habeeb01: isCurrent is used in this file `[valuationTab.scala.html](https://github.com/hmrc/voa-property-linking-frontend/pull/1164/files#diff-c783eeec6d6bd34ccb20386763640110d0f020d6f028bfabeaa22c8dbc105a3b)` I forgot to remove isPrevious (i didn't need it in the end)

PeterPerhac: image
i can only see it mentioned in `assessment.isCurrent` not in propertyLink. is `assessment` actually a property link?

PeterPerhac:
👍  looks good to me

DanSmith98:
LGTM 👍

reddy-yattapu:
👍

PR #1165, opened 2022-10-20
VTCCA-5179 updated ownership page error message when given date in the future

DanSmith98:


PeterPerhac:
👍  looks good to me

app/controllers/propertyLinking/ClaimPropertyOwnershipController.scala
```diff
@@ -115,7 +115,7 @@ object ClaimPropertyOwnership {
         "fromDate" -> mandatoryIfFalse(
           "interestedOnOrBefore",
           dmyDateAfterThreshold(earliestStartDate)
-            .verifying(Errors.dateMustBeInPast, d => d.isBefore(LocalDate.now))
+            .verifying("interestedOnOrBefore.error.dateInFuture", d => d.isBefore(LocalDate.now))
```
PeterPerhac, line 118: what's the difference between the dateInFuture error and the `dateMustBeInPast` ? sounds the same to me. Did we need to replace the error reference with a String here? I wonder if we're removing a reference to `Errors. dateMustBeInPast` here, then are there still other references to it from other places? (or can it be removed completely)

DanSmith98: Really I just changed the error string to correspond to the messages key because I thought it'd be easier to deal with that way.

PeterPerhac: ah, ok 👍

habeeb01:
LGTM 👍

PR #1166, opened 2022-10-25
Vtcca 5196 add link to dvr check case details

DanSmith98:


app/config/ApplicationConfig.scala
```diff
@@ -20,9 +20,9 @@ import java.util.Base64
 import com.google.inject.{Inject, Singleton}
 import play.api.Configuration
 
+import java.text.MessageFormat.format
```
PeterPerhac, line 23: looks like an unused import

app/views/dvr/dvrFiles.scala.html
```diff
@@ -74,56 +70,62 @@
...
         )
     )
     val challenge: TabItem = TabItem(
             id = Some("challenge-cases-tab"),
             label = messages("available.requestvaluation.challengesTab.title", challengeCases.size),
             panel = TabPanel(
-              content = HtmlContent(challengeCasesTabContent)
+              content = HtmlContent(
+                  challengeCasesDetailsTab(
+                      model.authorisationId,
```
PeterPerhac, line 114: same sort of thing here, picking all sorts of things out of the model, when you could just pass the model

app/views/dvr/tabs/checkCasesDetailsTab.scala.html
```diff
@@ -23,7 +23,20 @@
 @import java.time.LocalDate
 @import views.html.includes.helpIcon
 
-@(propertyLinkSubmissionId: String, isOwner: Boolean, uarn: Long, authorisationId: Long, assessmentRef: Long, listYear:String, checkCases: List[CaseDetails], downloadUrl: String, startCheckUrl: String)(govukButton: GovukButton, govukTable: GovukTable, govukDetails: GovukDetails, govukWarningText: GovukWarningText)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
+@this(govukButton: GovukButton, govukDetails: GovukDetails, govukTable: GovukTable, govukWarningText: GovukWarningText)
+
+@(
+  propertyLinkSubmissionId: String,
```
PeterPerhac, line 29: this tab could have its own model case class

test/controllers/detailedvaluationrequest/DvrControllerSpec.scala
```diff
@@ -416,6 +422,17 @@ class DvrControllerSpec extends VoaPropertyLinkingSpec {
...
         cssClasses should not include "govuk-tag--grey"
     }
+
+    Inspectors.forAll(checkCaseDetails) { details =>
+      checksTable
+        .getElementsContainingOwnText(details.caseReference)
+        .attr("href") shouldBe s"""/check-case/${details.caseReference}/summary?
+                                  |propertyLinkSubmissionId=${assessments.submissionId}&
+                                  |isOwner=true&
+                                  |valuationId=${assessments.assessments.headOption.fold(
```
PeterPerhac, line 432: here, it would be better if you separated out the `${assessments.assessments.headOption.fold(...` into a separate `val` before you dive into string interpolation. Then you could just fit it all on a single line I guess, and would not need the `.stripMargin.replaceAll("\n", "")`

app/config/GuiceModule.scala
```diff
@@ -69,9 +69,18 @@ class GuiceModule(
...
+    bindCheckFrontendEndpoints()
     bindVmvFrontendEndpoints()
   }
 
+  private def bindCheckFrontendEndpoints(): Unit =
+    bindEndpoints(
+      Map(
+        "check.summary.url" -> "resources.check.frontend.summary.path"
+      ),
+      servicesConfig.getString("business-rates-check.url")
```
PeterPerhac, line 81: this will get the check BACKEND service base url, no? https://github.com/hmrc/app-config-base/blob/main/voa-property-linking-frontend.conf#L77
I wouldn't be surprised if this worked fine locally but failed to start up in Dev/QA, can you double-check the config?

DanSmith98: Good thinking, I'll have a look at the config now.

DanSmith98:
@PeterPerhac I've moved the check case tab params to a model, and used the existing model for the other tabs you've mentioned.

PeterPerhac:
👍 looks good to me

PR #1186, opened 2022-12-05
VTCCA-5201 Scala 2.13 upgrade

zachhall1234:


test/utils/MultipartFormDataWritable.scala
```diff
@@ -57,10 +55,6 @@ object MultipartFormDataWritable {
   val singleton = Writeable[MultipartFormData[TemporaryFile]](
     transform = { form: MultipartFormData[TemporaryFile] =>
       formatDataParts(form.dataParts) ++
-        form.files.flatMap { file =>
```
DanSmith98, line 60: Do you think we can get rid of this whole object? I can't find anywhere it (or the implicit) is in use.

build.sbt
```diff
@@ -56,7 +52,7 @@ lazy val microservice = Project(appName, file("."))
     IntegrationTest / testGrouping := oneForkedJvmPerTest((IntegrationTest / definedTests).value),
     IntegrationTest / parallelExecution := false
   )
-  .settings(scalaVersion := "2.12.12")
+  .settings(scalaVersion := "2.13.8")
```
DanSmith98, line 55: Can we go straight to 2.13.10? If the platform has specified 2.13.8 then no worries.

zachhall1234: The guide mentioned 2.13.8 & the catalogue also did so I just went for that. I can have have a look if 2.13.10 is compatible with our dependencies.

DanSmith98: Ah, I see it in the guide screenshot. 2.13.10 came out after they published the guide and 2.13.9 was only two days before.

I think it's probably best to be on the newest available version, but have no issues with 2.13.8 if our libraries aren't compatible or something.

test/utils/MultipartFormDataWritable.scala
zachhall1234: Good spot, doesn't look like its being used. I've removed in the recent commit.

DanSmith98:
LGTM 👍

test/utils/MultipartFormDataWritable.scala
DanSmith98: I just checked the new commit, I think we can get rid of the whole file.

PeterPerhac:
👍  looks good to me

PR #1191, opened 2022-12-16
VTCCA-5201 uplift to latest bootstrap version to resolve http-verbs bug & targetJvm to jvm-11

zachhall1234:


DanSmith98:
LGTM 👍

PeterPerhac:
the PR has a title `uplift targetJvm to jvm-11` but it doesn't actually uplift jvm version?  Should this be closed? Why can't we uplift java version?

zachhall1234:
> the PR has a title `uplift targetJvm to jvm-11` but it doesn't actually uplift jvm version? Should this be closed? Why can't we uplift java version?

Apologies, me & Dan thought that the exceptions we was seeing in development/QA looked like it was struggling to find some Java 8 libraries, so initially created a hot fix with the targetJvm set to JVM-11 to see if this resolved the issue which it didn't. 

Turns on theres a bug in http-verbs for Scala 2.13, where the returned headers are case sensitive. This is fixed in the latest version of bootstrap-play (currently 7.11.0 ). I've deployed the hotfix to dev & it has resolved the MissingResponseHeader exceptions we was getting.

I'll update the PR description now. As for the TargetJvm i'm happy to add back in Java11 we was just unsure whether it would be best waiting till all services are building with Java11?

DanSmith98:
I'm wondering if setting the target to 11 was what caused Colin's service manager to not start locally when he didn't have 11 installed.

If that's the case, then this could cause problems for working on non-updated services, where the unit tests will sometimes fail on Java 11 due to changes with time precision and Locales. Currently, I can uninstall 11 and run on 8 if I'm working on a service which hasn't been updated, and service manager will still work fine. Targeting 11 will lose this flexibility.

The other thing to consider is that we can only target 11 after we do the 2.13 upgrades, so we could potentially be stuck for a while targeting different JVMs across services, even after the Java 11 update work has been done.

I'm open to targeting 11 here, but just wanted to flag that it could maybe cause problems with non-upgraded services.

zachhall1234:
After team discussion, we are going to prioritise Java11 updates for remaining services. TargetJvm uplift to Java11 isn't required for the Scala 2.13 as mentioned [here](https://jira.tools.tax.service.gov.uk/browse/VTCCA-5328) in this spikes comment.

PR #1198, opened 2023-01-13
VTCCA-5413 End date default for 2017 previous properties

reddy-yattapu:


zachhall1234:
LGTM 👍 PR builder seems to be failing though

PR #1223, opened 2023-06-22
VTCCA-5796 - Add new listYears field to getMyOrganisationAgents call

kabuumu:


zachhall1234:
LGTM 👍

PeterPerhac:
👍 looks good. Thanks for setting this up.

zachhall1234:
We probably need to update precommit alias to include the it:test.

Potentially in build-jobs too

MarkAKelly:
In the requirements we've got the addition of a feature switch to toggle whether the optional data is required:

>  On frontend, if feature switch is enabled then will throw exception if field is not present.

Is that meant to be added as part of this change or another time?

PR #1224, opened 2023-06-27
VTCCA-5825 Adding the new agent screens

joeyymoore:


it/base/ListYearsFeatureSwitchedISpecBase.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  lazy val wireMockServer: WireMockServer = new WireMockServer(wireMockConfig().port(mockPort))
+
+  val mockedMicroservices: Set[String] = Set(
+    "property-linking",
+    "business-rates-authorisation",
+    "auth"
+  )
+
+  val config: Map[String, String] = Map(
```
MarkAKelly, line 36: You probably don't need to duplicate the entire ISpecBase to change the config for the test you want.

You can try overriding the config val in the ISpec that needs to be modified with the values that you need for that test.

MarkAKelly:
LGTM

michjanowicz:
lgtm 👍

PR #1226, opened 2023-07-10
Vtcca 5798 update existing screens of agent journey

michjanowicz:
Passed branch testing 10/07

### DIFF
--- a/app/controllers/agent/ManageAgentController.scala
+++ b/app/controllers/agent/ManageAgentController.scala
@@ -159,35 +159,35 @@ class ManageAgentController @Inject()(
       ipPropertyLinksCount = propertyLinks.total
     } yield {
       agentToBeManagedOpt match {
-        case None => throw new NotFoundException("Agent is not available")
+        case None => None
         case Some(agent) =>
           if (featureSwitch.isAgentListYearsEnabled) {
             val ManageAgentOptionItemList =
-              (ipPropertyLinksCount, agentToBeManagedOpt) match {
-                case (0, Some(agent)) if agent.propertyCount == 0 =>
+              ipPropertyLinksCount match {
+                case 0 if agent.propertyCount == 0 =>
                   List(
                     ManageAgentOptionItem(ChangeRatingList),
                     ManageAgentOptionItem(RemoveFromYourAccount)
                   )
-                case (1, Some(agent)) if agent.propertyCount == 0 =>
+                case 1 if agent.propertyCount == 0 =>
                   List(
                     ManageAgentOptionItem(AssignToYourProperty),
                     ManageAgentOptionItem(ChangeRatingList),
                     ManageAgentOptionItem(RemoveFromYourAccount)
                   )
-                case (1, Some(agent)) if agent.propertyCount == 1 =>
+                case 1 if agent.propertyCount == 1 =>
                   List(
                     ManageAgentOptionItem(UnassignFromYourProperty),
                     ManageAgentOptionItem(ChangeRatingList)
                   )
-                case (numberOfPropertyLinks, Some(agent)) if numberOfPropertyLinks > 1 && agent.propertyCount == 0 =>
+                case numberOfPropertyLinks if numberOfPropertyLinks > 1 && agent.propertyCount == 0 =>
                   List(
                     ManageAgentOptionItem(AssignToAllProperties),
                     ManageAgentOptionItem(AssignToSomeProperties),
                     ManageAgentOptionItem(ChangeRatingList),
                     ManageAgentOptionItem(RemoveFromYourAccount),
                   )
-                case (numberOfPropertyLinks, Some(agent)) if numberOfPropertyLinks > agent.propertyCount =>
+                case numberOfPropertyLinks if numberOfPropertyLinks > agent.propertyCount =>
                   List(
                     ManageAgentOptionItem(AssignToAllProperties),
                     ManageAgentOptionItem(AssignToSomeProperties),
@@ -195,7 +195,7 @@ class ManageAgentController @Inject()(
                     ManageAgentOptionItem(UnassignFromSomeProperties),
                     ManageAgentOptionItem(ChangeRatingList)
                   )
-                case (numberOfPropertyLinks, Some(agent)) if numberOfPropertyLinks == agent.propertyCount =>
+                case numberOfPropertyLinks if numberOfPropertyLinks == agent.propertyCount =>
                   List(
                     ManageAgentOptionItem(UnassignFromAllProperties),
                     ManageAgentOptionItem(UnassignFromSomeProperties),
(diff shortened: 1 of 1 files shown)


app/controllers/agent/ManageAgentController.scala
```diff
@@ -84,6 +93,7 @@ class ManageAgentController @Inject()(
                             .getMyAgentPropertyLinks(agentCode, params, PaginationParams(1, 100, true))
         agentDetails <- agentRelationshipService.getAgentNameAndAddress(agentCode)
         myAgents     <- agentRelationshipService.getMyOrganisationAgents()
+        listYears = myAgents.agents.flatMap(_.listYears).headOption
```
kabuumu, line 96: This will return a sequence of list years that a user's agent can act upon, but only for the first agent in their list? Is that correct?

Looking at how you've used this value in the view:
```
    listYears.fold("Rating lists for this property are not available yet") { years =>
        years match {
            case Seq(listYear) => messages("propertyRepresentation.ratingList.p", listYear)
            case _ => messages("propertyRepresentation.ratingLists.p")
        }
    }
 ```
It will display "Rating lists for this property are not available yet" if your agents cannot act on any of your list years. I might be misunderstanding the requirement, but is this correct?

File: app/views/propertyrepresentation/manage/manageAgentProperties.scala.html [hunk 2]
```diff
@@ -14,52 +14,54 @@
...
-
 @(      ownerAuthResult: OwnerAuthResult,
         agentCode: Long,
-        agentDetails:Option[AgentDetails], backLink: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
+        agentDetails:Option[AgentDetails],
+        listYears: Option[Seq[String]],
+        backLink: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
 
+@ratingList = @{
+    listYears.fold("Rating lists for this property are not available yet") { years =>
```
kabuumu, line 31: Should this text be pulled from the messages file so that it can be translated?

michjanowicz: This is just a text for testing. I forgot to replace it with "". Sorry for that.

kabuumu: What do you mean just for testing sorry, is listYears being None not a valid scenario?

app/models/propertyrepresentation/ManageAgentOptions.scala
```diff
@@ -59,30 +65,55 @@ object ManageAgentOptions extends NamedEnumSupport[ManageAgentOptions] {
...
   val multiplePropertyLinksAgentAssignedToAllOptions = List(
     ManageAgentOptionItem(s"${UnassignFromAllProperties.name}"),
     ManageAgentOptionItem(s"${UnassignFromSomeProperties.name}")
   )
+  val multiplePropertyLinksAgentAssignedToAllOptionsWithRatingList = List(
+    ManageAgentOptionItem(s"${UnassignFromAllProperties.name}"),
+    ManageAgentOptionItem(s"${UnassignFromSomeProperties.name}"),
+    ManageAgentOptionItem(s"${ChangeRatingList.name}")
+  )
 }
```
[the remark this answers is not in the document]
michjanowicz: Thanks for comment. Do you think about something specific? I could use Enumeration to make it tidier... We will reduce options to 5 as soon as listYears is live as we got 5 valid scenarios of this screen in new version and 4 in the old one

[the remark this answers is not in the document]
michjanowicz: Is updated now :)

app/views/propertyrepresentation/manage/manageAgentProperties.scala.html
```diff
@@ -14,52 +14,52 @@
...
-
 @(      ownerAuthResult: OwnerAuthResult,
         agentCode: Long,
-        agentDetails:Option[AgentDetails], backLink: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
-
+        agentDetails:Option[AgentDetails],
+        listYears: Seq[String],
+        backLink: String)(implicit request: Request[_], messages: Messages, config: ApplicationConfig)
 
+@ratingList = @{
```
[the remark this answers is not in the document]
michjanowicz: Hi is updated now

app/controllers/agent/ManageAgentController.scala
```diff
@@ -142,59 +159,128 @@ class ManageAgentController @Inject()(
...
-          Some(
-            manageAgentView(
-              submitManageAgentForm,
-              ManageAgentOptions.multiplePropertyLinksNoAssignedAgentsOptions,
-              agent,
-              calculateBackLink(organisationsAgents, agent.representativeCode)
-            ))
+      if (featureSwitch.isAgentListYearsEnabled) {
+        val ManageAgentOptionItemList =
+          (ipPropertyLinksCount, agentToBeManagedOpt) match {
```
kabuumu, line 164: There's not a case here for when agentToBeManagedOpt is None. It is probably better to do that check above before the feature switch check, as if the user is on this page without an agent then nothing should work, regardless of feature switch or how many properties they have.

it/connectors/propertyLinking/ManageAgentISpec.scala
```diff
@@ -0,0 +1,335 @@
+package connectors.propertyLinking
```
kabuumu, line 1: This is a controller test, not a connector one, isn't it?

kabuumu (changes requested):

it/connectors/propertyLinking/ManageAgentISpec.scala
michjanowicz: Yes.  I'll  place it (and the 2 others) in the controller package

app/controllers/agent/ManageAgentController.scala
michjanowicz: Code has been updated, branch rebased.

app/controllers/agent/ManageAgentController.scala
```diff
@@ -142,59 +158,128 @@ class ManageAgentController @Inject()(
...
+              }
+            Some(
+              manageAgentView(
+                submitManageAgentForm,
+                ManageAgentOptionItemList,
+                agent,
+                calculateBackLink(organisationsAgents, agent.representativeCode)
+              ))
+          } else {
+            (ipPropertyLinksCount, agentToBeManagedOpt) match {
```
kabuumu, line 213: `agent` is already available here taken out of the option so you don't need to match on it again.

app/controllers/agent/ManageAgentController.scala
```diff
@@ -142,59 +158,128 @@ class ManageAgentController @Inject()(
...
-              calculateBackLink(organisationsAgents, agent.representativeCode)))
-        case (1, Some(agent)) if agent.propertyCount == 1 =>
-          //IP has one property link and agent is assigned to that property
-          Some(
-            unassignAgentFromPropertyView(
-              submitAgentAppointmentRequest,
-              agent,
-              calculateBackLink(organisationsAgents, agent.representativeCode)))
+      agentToBeManagedOpt match {
+        case None => throw new NotFoundException("Agent is not available")
```
kabuumu, line 162: Does this need to throw an exception here, or can it return None, as the method returns an optional view and the NotFound is handled above?

app/controllers/agent/ManageAgentController.scala
michjanowicz: Should I do `case None => None` or case _ => None ??

PR #1227, opened 2023-07-14
Updates to revoke properties aswell as appoint

joeyymoore:


it/connectors/propertyLinking/AreYouSureISpec.scala
```diff
@@ -1,8 +1,9 @@
 package connectors.propertyLinking
```
kabuumu, line 1: Is this a controller test, not a connector one?

app/controllers/propertyLinking/AreYouSureController.scala
```diff
@@ -65,21 +65,42 @@ class AreYouSureController @Inject()(
 
   def submitRatingListYears(chosenListYear: String): Action[AnyContent] = authenticated.async { implicit request =>
     if (featureSwitch.isAgentListYearsEnabled) {
-      manageAgentSessionRepository.get[AgentSummary].map {
+      manageAgentSessionRepository.get[AgentSummary].flatMap {
```
kabuumu, line 68: Because this code is mostly reused across this and the AreYouSureMultipleController, would it be better moving into a service class?

kabuumu (approved):

PR #1228, opened 2023-07-20
VTCCA-5871 appoint agents defaulted 2017&2023, one connector call to …

zachhall1234:
…one BE endpoint


joeyymoore:
LGTM 👍 manually tested and AT's pass when this is merged with Michals changes

PR #1230, opened 2023-07-24
VTCCA-5798 bugs found during testing fixed

zachhall1234:


joeyymoore:
ATs are now passing so LGTM 👍

PR #1234, opened 2023-08-08
VTCCA-5918 add call appoint all/some properties to check agent list y…

zachhall1234:
…ear access


joeyymoore:
for your changes, youve made changes for appoint all and appoint some right? do you need to add any tests for the appoint some flow?

zachhall1234:
> for your changes, youve made changes for appoint all and appoint some right? do you need to add any tests for the appoint some flow?

Good spot, added these tests in!

joeyymoore (approved):

PR #1235, opened 2023-08-09
VTCCA-5939 Update content to say we may contact users using contact details

joeyymoore:


MarkAKelly:
Looks good to me boss

PR #1245, opened 2023-08-22
VTCCA-5986 passed round fromManageAgentJourney flag around all functi…

zachhall1234:
…ons on the appoint properties page


WillBlakeney:
LGTM 👍  I haven't approved in case somebody with more dev experience wants to take a look as well

PR #1250, opened 2023-11-13
VTCCA-6248 update bootstrap to include new CSP filter

carterd888:
updated bootstrap for CSP

acceptance tests pass, copy of logs included
Screenshot 2023-11-13 at 10 55 05

### DIFF
--- a/app/views/dvr/dvrFiles.scala.html
+++ b/app/views/dvr/dvrFiles.scala.html
@@ -154,7 +154,7 @@
     <br/>
 
     @for(activeTabId <- model.activeTabId) {
-        <script type="text/javascript">
+        <script CSPNonce.attr type="text/javascript">
         document.addEventListener("DOMContentLoaded", function() {
             location.hash = "@activeTabId";
             @if(startCheckForm.hasErrors){
--- a/app/views/dvr/dvrFiles.scala.html
+++ b/app/views/dvr/dvrFiles.scala.html
@@ -20,6 +20,7 @@
 @import models.dvr.cases.check.projection.CaseDetails
 @import models.dvr.cases.check.StartCheckForm
 @import models.properties.AllowedAction.CHECK
+@import views.html.helper.CSPNonce
 
 @this(
   mainLayout: views.html.mainLayout,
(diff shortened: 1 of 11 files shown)


build.sbt
```diff
@@ -123,7 +123,7 @@ lazy val compileDependencies = Seq(
   "com.google.guava"     % "guava"                          % "18.0",
   "org.typelevel"        %% "cats-core"                     % "2.6.1",
   "uk.gov.hmrc"          %% "bootstrap-frontend-play-28"    % bootstrapPlayVersion,
-  "uk.gov.hmrc"          %% "play-frontend-hmrc"            % "5.5.0-play-28",
```
WillBlakeney, line 126: play-frontend-hmrc 7.0.0 is from March. Could we go to the latest (7.27.0)?

build.sbt
```diff
@@ -123,7 +123,7 @@ lazy val compileDependencies = Seq(
...
+  "uk.gov.hmrc"          %% "play-frontend-hmrc"            % "7.0.0-play-28",
```
WillBlakeney, line 126: play-frontend-hmrc 7.0.0 is from March. Could we go to the latest (7.27.0)?

build.sbt
```diff
@@ -113,7 +113,7 @@ scalacOptions += "-Wconf:src=target/.*:s"
 
 scalacOptions ++= Seq("-unchecked", "-deprecation", "-feature")
 
-val bootstrapPlayVersion = "7.19.0"
+val bootstrapPlayVersion = "7.22.0"
```
WillBlakeney, line 116: Pretty minor thing, but 7.23.0 was released last week so we should probably use that

WillBlakeney (approved):
LGTM 👍

app/views/dvr/dvrFiles.scala.html
```diff
@@ -154,7 +155,7 @@ <h1 id="assessment-address" class="govuk-heading-l">@address</h1>
     <br/>
 
     @for(activeTabId <- model.activeTabId) {
-        <script type="text/javascript">
+        <script @{CSPNonce.attr} type="text/javascript">
         document.addEventListener("DOMContentLoaded", function() {
             location.hash = "@activeTabId";
             @if(startCheckForm.hasErrors){
```
WillBlakeney, line 161: I've just double checked the Confluence page and I don't think we can use inline javascript anymore. I think this needs to be changed

WillBlakeney (approved):
We are removing inline javascript in https://jira.tools.tax.service.gov.uk/browse/VTCCA-6271, so this is fine to be squashed and merged.

PR #1260, opened 2024-01-11
VTCCA-6218 add interface tests for showRemoveAgentFromIpOrganisation

WillBlakeney:


app/controllers/agent/ManageAgentController.scala
```diff
@@ -425,7 +425,8 @@ class ManageAgentController @Inject()(
           form = submitAgentAppointmentRequest,
           agentCode = agent.representativeCode,
           agentName = agent.name,
-          backLink = config.dashboardUrl("home")))
+          backLink = controllers.agent.routes.ManageAgentController.showManageAgent.url
+        ))
```
WillBlakeney, line 429: This is halfway through a journey, so I don't understand why the back link was going home. I have updated it to go back to the previous page

zachhall1234 (approved):

PR #1269, opened 2024-01-24
VTCCA-6216 Interface tests for the /my-organisation/confirm-appoint-agent page

joeyymoore:


it/controllers/agentAppointment/ConfirmAgentAppointControllerFSOnISpec.scala
```diff
@@ -18,110 +18,365 @@ import java.util.UUID
...
+      managingPropertyChoice = ChooseFromList.name,
       agentAddress = "An Address",
       backLink = None,
       totalPropertySelectionSize = 2,
       propertySelectedSize = 2,
-      bothRatingLists = bothRatingList,
-      specificRatingList = specificRatingList
+      singleProperty = true,
+      bothRatingLists = bothRatingsListChoice,
+      specificRatingList = specificRatingsListChoice
```
WillBlakeney, line 379: All the tests are when the choice is ChooseFromList & assigning all 2 properties to the agent. I know the content shouldn't change, but we run this test every time
```
s"doesn't have the text on the screen for the assigned properties" in {
        document.select(assignedToPropertiesSelector).size() shouldBe 0
}
```

It might be worth testing assign to all/some/one/none to make sure the different variations of assignedPropertiesText don't show up. I don't think we need separate tests, adding it to the existing is fine e.g all & 2023, some & 2017, one property & both list years then one additional test for assigned to none

it/controllers/agentAppointment/ConfirmAgentAppointControllerISpec.scala
```diff
@@ -17,136 +17,371 @@ import java.util.UUID
...
+        document.select(thisAgentCanSelector).text shouldBe thisAgentCanText
+      }
+
+      s"has a bullet point on the screen of $addPropertiesText" in {
+        document.select(addPropertiesSelector).text shouldBe addPropertiesText
+      }
+
+      s"doesn't not have a bullet point on the screen for the ratings list years" in {
+        document.select(ratingsListSelector).size() shouldBe 0
+      }
```
WillBlakeney, line 102: Pretty minor, but throughout this spec you've got a copy and paste error of `doesn't not have` which as a double negative sounds like it **_does_** have but you're testing that it **_doesn't_** have

WillBlakeney:
When the feature switch is off, could you add a test for assigning to ‘none’. It shouldn’t have any "#assigned-to" text but I think everything else is the same

it/controllers/agentAppointment/ConfirmAgentAppointControllerISpec.scala
joeyymoore: Its not minor, thats horrendous 😅 i'll update this

joeyymoore:
> When the feature switch is off, could you add a test for assigning to ‘none’. It shouldn’t have any "#assigned-to" text but I think everything else is the same

I didnt realise this was a valid scenario, so will add a test for it now

it/controllers/agentAppointment/ConfirmAgentAppointControllerFSOnISpec.scala
WillBlakeney: Okay cool, that's fine then 👍

WillBlakeney (approved):
LGTM 👍

PR #1271, opened 2024-01-30
VTCCA-6204: Interface tests for /my-organisation/claim/property-links/ownership

joeyymoore:


joeyymoore:
Issues that will be covered in a separate ticket

- 2 h1s on the page
- Multiple error messages pop up if you cause different types of errors simultaneously 
- Singular errors don’t pop up for each input field
- Invalid date error in summary doesn’t match the one above the field

- Doesn’t translate owner/occupier to welsh on the error messages
- Error messages missing a “the” e.g. “Date you became owner” should be “Date you became the owner”

it/controllers/propertyLinking/ClaimPropertyOwnershipISpec.scala
```diff
@@ -0,0 +1,1493 @@
...
+  val captionText = "Add a property"
+  val headingText = "When you became the owner or occupier of the property"
+  val headingTextAgent = "When your client became the owner or occupier of the property"
+  val iHaveOwnedText = "I have owned or occupied the property on more than one occasion"
+  val iHaveOwnedTextAgent = "My client has owned or occupied the property on more than one occasion"
+  val youCanAddText = "You can add the property for each period you were connected to it, but you need to do it one period at a time. However, if you want to talk to the Valuation Office Agency about the valuation, you will need to select the right valuation period."
+  val youCanAddTextAgent = "You can add the property for each period your client was connected to it, but you need to do it one period at a time. However, if you want to talk to the Valuation Office Agency about the valuation, you will need to select the right valuation period."
+  val onWhatDateText = "On what date did you become the owner or occupier?"
+  val onWhatDateTextAgent = "On what date did your client become the owner or occupier?"
+  val forExampleText = "For example, 1 4 2017"
```
WillBlakeney, line 44: I think this forExampleText changes depending on if the property is English or Welsh. So for English its 1 4 2017 and for Welsh it's 1 4 2023 - It might be worth adding a test for that

it/controllers/propertyLinking/ClaimPropertyOwnershipISpec.scala
```diff
@@ -0,0 +1,1493 @@
...
+        .withHttpHeaders(HeaderNames.COOKIE -> "sessionId", "Csrf-Token" -> "nocheck")
+        .post(body = requestBody)
+    )
+  }
+
+  private def genericUserPageStuff(day: String, month: String, year: String, relationshipChoice: String, endDate: Option[PropertyOccupancy] = None): Unit = {
+
+    lazy val res = submitOwnership(day = day, month = month, year = year, relationship = relationshipChoice, endDate = endDate)
+
+    lazy val document = Jsoup.parse(res.body)
```
WillBlakeney, line 1257: I spent ages trying to do something like this before and gave up. My only concern with this way is that we're making a new request each time, even though we just made a request to test the error messages, which could add some extra time to the tests.

I think it would be good to discuss this and hear other people's views on it because I honestly don't know what's best

it/controllers/propertyLinking/ClaimPropertyOwnershipISpec.scala
```diff
@@ -0,0 +1,1493 @@
...
+    await(
+      ws.url(s"http://localhost:$port/business-rates-property-linking/my-organisation/claim/property-links/ownership")
+        .withCookies(languageCookie(language), getSessionCookie(testSessionId))
+        .withFollowRedirects(follow = false)
+        .withHttpHeaders(HeaderNames.COOKIE -> "sessionId", "Csrf-Token" -> "nocheck")
+        .post(body = requestBody)
+    )
+  }
+
+  private def genericUserPageStuff(day: String, month: String, year: String, relationshipChoice: String, endDate: Option[PropertyOccupancy] = None): Unit = {
```
WillBlakeney, line 1253: I think `genericUserPageTests`, `genericAgentPageTests` etc is clearer to understand what it's doing than `genericUserPageStuff`, `genericAgentPageStuff` etc

WillBlakeney:
I think there's some tests that could be removed from `ClaimPropertyOwnershipControllerSpec` now

it/controllers/propertyLinking/ClaimPropertyOwnershipISpec.scala
joeyymoore: I went through it with Mark, and when i ran the test with it doing the request once/twice, the difference was minimal. (when i ran it with it not doing double requests, on one occasion the tests took longer 😅)

joeyymoore:
> I think there's some tests that could be removed from `ClaimPropertyOwnershipControllerSpec` now

Cool will take a look

it/controllers/propertyLinking/ClaimPropertyOwnershipISpec.scala
joeyymoore: ah yeah i meant to rename that something that wasn't stuff 😅

it/controllers/propertyLinking/ClaimPropertyOwnershipISpec.scala
joeyymoore: Isnt it set by whatever earliestStartDate is set in session, so even though English/Welsh is different, it'll only ever display whats there i.e. i could manually change it to 1 1 2010 and it'll display that

WillBlakeney: Ah okay cool 👍

WillBlakeney (approved):

PR #1274, opened 2024-02-06
VTCCA-6207: Interface tests for /my-organisation/claim/property-links/summary page

joeyymoore:


WillBlakeney:
One small comment, the name DeclarationControllerShowISpec implies it only covers the show method, but it also covers submit

WillBlakeney (approved):

PR #1276, opened 2024-02-08
VTCCA-6492 Updated the address line validation to be max 30 chars

joeyymoore:


app/form/Mappings.scala
```diff
@@ -48,10 +48,12 @@ object Mappings extends DateMappings {
 
   val addressMapping: Mapping[Address] = mapping(
     "addressId" -> addressId,
-    "line1"     -> onlyIf(addressEnteredManually, text.verifying(nonEmpty, maxLength(36)))(""),
-    "line2"     -> text(maxLength = 36),
-    "line3"     -> text(maxLength = 36),
-    "line4"     -> text(maxLength = 36),
+    "line1" -> onlyIf(
+      addressEnteredManually,
+      text.verifying(nonEmpty, maxLength(30, "enrolment.address.maxLengthError")))(""),
```
WillBlakeney, line 53: I think line one always has a limit of 80 characters according to Mahmed's comment on [VTCCA-6394](https://jira.tools.tax.service.gov.uk/browse/VTCCA-6394)

WillBlakeney:
When the user gets an error, the 'Enter address manually' link disappears. This means the user can only see the 'Postcode' field and not the error they actually received. I'm not sure if this is the scope creeping too much, but do you think we could make it so that the user can actually see their error?

WillBlakeney:
Also, just want to check a ticket has been raised to fix the other fields on the page, add an error summary, fix changing the language etc etc?

joeyymoore:
> Also, just want to check a ticket has been raised to fix the other fields on the page, add an error summary, fix changing the language etc etc?

Fix the other fields yes, fix the other stuff no. I can whip one up but theres so much wrong with it 😅

joeyymoore:
> When the user gets an error, the 'Enter address manually' link disappears. This means the user can only see the 'Postcode' field and not the error they actually received. I'm not sure if this is the scope creeping too much, but do you think we could make it so that the user can actually see their error?

This was on my list of things that are wrong with the page, because of how the page works i dont think its a straight forward fix (like most other pages would be) so will need a separate ticket

app/form/Mappings.scala
joeyymoore: okay i'll update the first line to be 80, and the corresponding error message too

WillBlakeney:
Wow, it's annoying you can't make a thread with normal comments on a PR. But no worries 👍

WillBlakeney:
rebuild

WillBlakeney (approved):

PR #1277, opened 2024-02-15
VTCCA-6132 implement govuk file upload component 

zachhall1234:


app/controllers/propertyLinking/DeclarationController.scala
```diff
@@ -72,18 +72,17 @@ class DeclarationController @Inject()(
...
+          case (RatesBillType, Owner | OwnerOccupier | Occupier)        => EvidenceChoices.RATES_BILL
+          case (ServiceCharge, Owner | OwnerOccupier | Occupier)        => EvidenceChoices.SERVICE_CHARGE
+          case (StampDutyLandTaxForm, Owner | OwnerOccupier | Occupier) => EvidenceChoices.STAMP_DUTY
+          case (LandRegistryTitle, Owner | OwnerOccupier | Occupier)    => EvidenceChoices.LAND_REGISTRY
+          case (WaterRateDemand, Owner | OwnerOccupier | Occupier)      => EvidenceChoices.WATER_RATE
+          case (OtherUtilityBill, Owner | OwnerOccupier | Occupier)     => EvidenceChoices.UTILITY_RATE
+          case (Lease, Occupier | Owner | OwnerOccupier | Occupier)     => EvidenceChoices.LEASE
+          case (License, Occupier | Owner | OwnerOccupier | Occupier)   => EvidenceChoices.LICENSE
+          case (_, Owner | OwnerOccupier | Occupier)                    => EvidenceChoices.OTHER
+          case (_, Occupier)                                            => EvidenceChoices.NO_LEASE_OR_LICENSE
```
joeyymoore, line 85: this case (_, Occupier) will never happen anymore right? now that youve added the Occupier to the above case statement, so its not needed?

app/views/propertyLinking/upload_result.scala.html
```diff
@@ -0,0 +1,160 @@
...
+        case (Some(_), Lease) => {
+            messages("uploadEvidence.lease.client.title")
+        }
+        case (Some(_), License) => {
+            messages("uploadEvidence.license.client.title")
+        }
+        case (Some(_), RatesBillType) => {
+            messages("uploadRatesBill.show.client.title")
+        }
+        case (Some(_), _) => {
```
joeyymoore, line 65: is there a client equivalent for this? or is it the same as whats displayed for case (None, _)? Could probably just do a case (_,_) at the end if so

app/views/propertyLinking/upload_result.scala.html
```diff
@@ -0,0 +1,160 @@
...
+        case (None, _) => {
+            messages(s"uploadEvidence.$evidenceType.title")
+        }
+    }
+}
+
+@tag(status: FileStatus) = @{
+    status match {
+        case _@FileStatus.UPLOADING => govukTag(Tag(content = Text(messages("fileUploadResult.uploading")), classes = "govuk-tag--yellow", attributes = Map("id" -> "uploadStatus")))
+        case _@FileStatus.READY => govukTag(Tag(content = Text(messages("fileUploadResult.uploaded")), classes = "govuk-tag--green", attributes = Map("id" -> "uploadStatus")))
```
joeyymoore, line 86: is it right that the READY status maps to fileUploadResult.uploaded?

conf/logback.xml
```diff
@@ -12,13 +12,13 @@
...
   <logger name="voa.submission" level="${logger.voa.submission:-DEBUG}"/>
 
-  <logger name="com.google.inject" level="WARN"/>
-  <logger name="reactivemongo" level="WARN"/>
-  <logger name="play" level="WARN"/>
-  <logger name="akka" level="WARN"/>
-  <logger name="uk.gov.hmrc.audit" level="WARN"/>
-  <logger name="auditCounter" level="WARN"/>
-  <logger name="uk.gov.hmrc.play.bootstrap" level="WARN"/>
+<!--  <logger name="com.google.inject" level="WARN"/>-->
```
joeyymoore, line 15: did you mean to leave this and the below commented out?

it/controllers/UploadResultControllerISpec.scala
```diff
@@ -0,0 +1,556 @@
...
+import play.api.test.Helpers._
+import repositories.PropertyLinkingSessionRepository
+import uk.gov.hmrc.http.{HeaderCarrier, SessionId}
+
+import java.time.LocalDate
+import java.util.UUID
+
+class UploadResultControllerISpec extends ISpecBase with HtmlComponentHelpers {
+
+  //  TODO: Add in welsh translations
```
joeyymoore, line 23: still need this TODO?

app/views/propertyLinking/upload_result.scala.html
zachhall1234: Yep, one of the issues with the design of the ticket is that we display uploading/uploaded which aren't accurate to whats happening but yes in this case Ready = Uploaded

it/controllers/UploadResultControllerISpec.scala
zachhall1234: I do not shall revert now cheers

app/views/propertyLinking/upload_result.scala.html
zachhall1234: Good spot, It appears to be the same, if im honest this is a straight copy from the upload.scala however I wonder if we should have content specific for clients for anything that falls in here-> case (Some(_), _)

joeyymoore (approved):

PR #1288, opened 2024-03-27
VTCCA-6586 Moving the dvr draft check creation to voa-pl-frontend from check-frontend

joeyymoore:


app/connectors/check/CheckConnector.scala
```diff
@@ -0,0 +1,113 @@
...
+      dvrCheck.map(d => s"dvrCheck=$d"),
+      rateableValueTooHigh.map(r => s"rateableValueTooHigh=$r")
+    ).flatten.mkString("?", "&", "")
+
+  private case class CheckIdWrapper(id: CheckId)
+
+  private object CheckIdWrapper extends PartialCheckFormats {
+    implicit val format: Reads[CheckIdWrapper] = Json.reads[CheckIdWrapper]
+  }
+
```
zachhall1234, line 112: Happy to be told otherwise but this could potentially be refactored to make use of the BusinessRatesCheckService layer & possibly move CheckIdWrapper into its own model?

zachhall1234 (approved):

PR #1306, opened 2024-10-09
VTCCA-7028 dependencies update

zachhall1234:


build.sbt
```diff
@@ -127,17 +127,17 @@ lazy val compileDependencies = Seq(
   "com.google.guava"     % "guava"                                  % "18.0",
```
joeyymoore, line 127: can this be bumped forward? latest version is like 33.3.1 (not sure which one you'd choose though but the latest none jre/android one is 23.0 anyway)

zachhall1234: Tbh I saw so many vulnerabilities on here that I thought best stick with the ones with high usage/what we know is working but I can bump it up & see what happens

zachhall1234: Turns out its not being used anymore

joeyymoore (approved):

PR #1318, opened 2025-03-06
VTCCA-7427: Updated content in registration journey

joenjones07:


app/form/Mappings.scala
```diff
@@ -36,6 +36,9 @@ import scala.util.matching.Regex
 
 object Mappings extends DateMappings {
 
+  val businessName: Mapping[String] =
```
zachhall1234, line 39: Does this still check that text is non empty? & can we add tests for this error please? Apologies if I missed this & its already done, I couldn't see it.

PR #1348, opened 2025-08-06
VTCCA-7810: Fixed language bug

joenjones07:


app/controllers/propertyLinking/UploadController.scala
```diff
@@ -177,7 +177,7 @@ class UploadController @Inject() (
         )
     }
 
-  def continue: Action[AnyContent] =
```
joeyymoore, line 180: can you / have you ran the acceptance tests against this change? they fail for me every time against the property linking block of tests...but manually its all fine. If they run fine for you it could be a joeys machine issue

PR #1390, opened 2026-08-28
VTCCA-8585 Add ObjectStorePending attachment state

kevin-co-hmrc:


app/models/attachment/AttachmentState.scala
```diff
@@ -45,6 +46,7 @@ object AttachmentState {
         case JsString("ScanReceived")              => JsSuccess(ScanReceived)
         case JsString("MetadataPending")           => JsSuccess(MetadataPending)
         case JsString("MetadataReceived")          => JsSuccess(MetadataReceived)
+        case JsString("ObjectStorePending")        => JsSuccess(ObjectStorePending)
```
joenjones07: Happy with the changes, could you amend the JsError on the Case _ in this match section to have a better message containing what attachment state it got. (This is my mistake just saves having a separate PR for it)

WillBlakeney (approved):

